PR #8, opened 2020-03-10
Create api definition activies#4

EliuX:
This PR creates an API definition to manage activities.

### DIFF
--- a/time_tracker_api/__init__.py
+++ b/time_tracker_api/__init__.py
@@ -9,6 +9,6 @@ def create_app():
     return flask_app
 
 
-def init_app(app):
+def init_app(app: Flask):
     from .api import api
     api.init_app(app)
--- a/time_tracker_api/__init__.py
+++ b/time_tracker_api/__init__.py
@@ -1,12 +1,14 @@
 from flask import Flask
-from flask_restplus import Api
 
 
 def create_app():
     flask_app = Flask(__name__)
-    app = Api(app=flask_app)
 
-    from time_tracker_api.projects.projects_endpoints import ns as projects_ns
-    app.add_namespace(projects_ns)
+    init_app(flask_app)
 
     return flask_app
+
+
+def init_app(app):
+    from .api import api
+    api.init_app(app)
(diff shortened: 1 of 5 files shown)


time_tracker_api/__init__.py
```diff
@@ -1,12 +1,14 @@
...
-    app = Api(app=flask_app)
 
-    from time_tracker_api.projects.projects_endpoints import ns as projects_ns
-    app.add_namespace(projects_ns)
+    init_app(flask_app)
 
     return flask_app
+
+
+def init_app(app):
```
Angeluz-07: Could we use type hints here  https://docs.python.org/3/library/typing.html ?

time_tracker_api/api.py
```diff
@@ -0,0 +1,12 @@
+
+from flask_restplus import Api
+
+api = Api(version='1.0.1', title="TimeTracker API",
+          description="API for the TimeTracker project")
+
+# APIs
+from time_tracker_api.projects import projects_api
```
Angeluz-07: Shouldn't we put all the imports at the top? 

On the other hand, the import `from time_tracker_api.projects import projects_api` from what I understand, you import the file `projects_api` and use it to reference functions inside of it. But It took me some minutes to figure it out. I think it would more easy to grasp something like `from time_tracker_api.projects.projects_api import ns as projects_ns`. But feel free to argument why your approach.

EliuX: The thing is that when I see the variable projects_ns it looks like an alias but it is not clear where it comes from, also it makes that import line bigger. When I see projects_api.ns, I understand that it comes from projects_api and that the variable it references is `ns`, there is no possible abstract interpretation.

EliuX: I did the import in the same part where I added that namespace because it seemed a good idea to have them together in case I wanted to remove or comment the inclusion of a particular namespace.

time_tracker_api/__init__.py
EliuX: Makes sense, I will do it.

PR #9, opened 2020-03-10
Create API definition for time entries

EliuX:
Creates the API definition for Time Entries, which are those created every time a user starts a new activity.

### DIFF
--- a/time_tracker_api/activities/activities_api.py
+++ b/time_tracker_api/activities/activities_api.py
@@ -4,28 +4,56 @@ ns = Namespace('activities', description='API for activities')
 
 # Activity Model
 activity = ns.model('Activity', {
-    'id': fields.String(readOnly=True, required=True, title='Identifier',
-                         description='The unique id of the activity'),
-    'name': fields.String(required=True, title='Name', max_length=50,
-                          description='Canonical name of the activity'),
-    'description': fields.String(title='Description',
-                                 description='Comments about the activity'),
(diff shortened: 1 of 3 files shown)


time_tracker_api/activities/activities_api.py
```diff
@@ -4,28 +4,72 @@
...
+        required=True,
+        title='Name',
+        max_length=50,
+        description='Canonical name of the activity',
+    ),
+    'description': fields.String(
+        title='Description',
+        description='Comments about the activity',
+    ),
+
```
Angeluz-07: You have a space here.

time_tracker_api/time_entries/time_entry_api.py
```diff
@@ -0,0 +1,132 @@
+from flask_restplus import fields, Resource, Namespace
+
+ns = Namespace('time_entries', description='API for time entries')
```
Angeluz-07: You may want to change `time_entries` to `time-entries` as recommendation from https://restfulapi.net/resource-naming/

time_tracker_api/activities/activities_api.py
EliuX: Refresh it, please :)

time_tracker_api/time_entries/time_entry_api.py
EliuX: True, my bad.

PR #12, opened 2020-03-11
Create specification for technologies

Angeluz-07:


time_tracker_api/technologies/technologies_api.py
```diff
@@ -0,0 +1,83 @@
...
+        max_length=50,
+        description='Name of the technology'
+    ),
+    'created_at': fields.Date(
+        title='Date',
+        description='Date of creation'
+    ),
+    'created_by': fields.String(
+        required=True,
+        title='Type',
```
EliuX: Copy & paste error. Put "Creator" as title

time_tracker_api/technologies/technologies_api.py
```diff
@@ -0,0 +1,83 @@
...
+        description='The tenant this technology belongs to'
+    ),
+    'name': fields.String(
+        required=True,
+        title='Name',
+        max_length=50,
+        description='Name of the technology'
+    ),
+    'created_at': fields.Date(
+        title='Date',
```
EliuX: I would like that these kinds of fields are spread equally in the rest of the models so `Date` is a little ambiguous, it could be confused with the start date of the time entries for example. I would leave as the title "Created" and the description is great.

time_tracker_api/technologies/technologies_api.py
```diff
@@ -0,0 +1,83 @@
...
+        description='Date of creation'
+    ),
+    'created_by': fields.String(
+        required=True,
+        title='Type',
+        max_length=30,
```
EliuX: This field will be an id so let's keep it as 64 of length.

time_tracker_api/technologies/technologies_api.py
```diff
@@ -0,0 +1,83 @@
...
+        description='Id of user who first added this technology',
+    ),
```
EliuX: Also, let's move these kinds of fields that are generated after creation, i.e. created_at, created_by, tenant_id, etc., to a model called `technology_response`, this way you will not require them when the technology is being created but this has to be autogenerated.

time_tracker_api/technologies/technologies_api.py
```diff
@@ -0,0 +1,83 @@
...
+})
+
+
+@ns.route('')
+class Technologies(Resource):
+    @ns.doc('list_technologies')
+    @ns.marshal_list_with(technology, code=200)
```
EliuX: Then in the `@ns.marshal_list_with` you will use `technology_response`.

time_tracker_api/technologies/technologies_api.py
EliuX: Those would be those ReadOnly fields

PR #15, opened 2020-03-11
Add CLI commands to generate API definition

EliuX:
This PR makes possible to run CLI commands:
- gen_swagger_json: Exports swagger specifications in JSON format, namely swagger.json
- gen_postman_collection: Generates a Postman Collection for the API

Check out the README for better information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,16 +52,16 @@ a link to the swagger.json with the definition of the api.
 
 ## CLI
 
-There are available commands aware of the API that can be verify useful for you. You
+There are available commands, aware of the API, that can be very helpful to you. You
 can check them out by running
 
-```angular2
+```
 python cli.py
 ```
 
 If you want to run an specific command, e.g. `gen_swagger_json`, specify it as a param
 as well as its correspondent options.
 
-```angular2
+```
 python cli.py gen_swagger_json -f ~/Downloads/swagger.json
 ```
--- a/README.md
+++ b/README.md
@@ -48,3 +48,20 @@ automatically [pip](https://pip.pypa.io/en/stable/) as well.
 
 - Open `http://127.0.0.1:5000/` in a browser. You will find in the presented UI 
 a link to the swagger.json with the definition of the api.
+
+
+## CLI
+
+There are available commands aware of the API that can be verify useful for you. You
+can check them out by running
+
+```angular2
+python cli.py
+```
+
+If you want to run an specific command, e.g. `gen_swagger_json`, specify it as a param
+as well as its correspondent options.
+
+```angular2
+python cli.py gen_swagger_json -f ~/Downloads/swagger.json
+```
--- a/cli.py
+++ b/cli.py
@@ -7,8 +7,9 @@ print("****************")
 import os
 
 from flask import json
-from time_tracker_api import create_app
 from flask_script import Manager
+
+from time_tracker_api import create_app
 from time_tracker_api.api import api
 
 app = create_app()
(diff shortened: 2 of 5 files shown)


cli.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@cli_manager.command
+@cli_manager.option('-f', '--filename', dest='filename', help='Path of the postman collection file.'
+                                         'By default it is timetracker-api-postman-collection.json')
+def gen_postman_collection(filename='timetracker-api-postman-collection.json'):
+    """ Generates a Postman collection for the API """
+    data = api.as_postman(urlvars=False, swagger=True)
+    postman_collection_json_data = json.dumps(data)
+    parsed_json = postman_collection_json_data.replace("http://localhost", '{{timetracker_api_host}}')
```
fabidick22: It would be great if we can send `http://localhost` as a parameter to the `cli.py` file. I think by default it can be localhost.

README.md
```diff
@@ -47,4 +47,21 @@ automatically [pip](https://pip.pypa.io/en/stable/) as well.
...
\ No newline at end of file
+a link to the swagger.json with the definition of the api.
+
+
+## CLI
+
+There are available commands aware of the API that can be verify useful for you. You
+can check them out by running
+
+```angular2
```
fabidick22: :eyes:

README.md
```diff
@@ -47,4 +47,21 @@ automatically [pip](https://pip.pypa.io/en/stable/) as well.
...
+python cli.py
+```
+
+If you want to run an specific command, e.g. `gen_swagger_json`, specify it as a param
+as well as its correspondent options.
+
+```angular2
```
fabidick22: :eyes:

cli.py
```diff
@@ -0,0 +1,52 @@
...
+    save_data(parsed_json, filename)
+
+
+def save_data(data, filename):
```
Angeluz-07: Could we have type hints here?

cli.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
+print("****************")
```
Angeluz-07: Is it necessary to have these prints before the import section ?

README.md
```diff
@@ -47,4 +47,21 @@ automatically [pip](https://pip.pypa.io/en/stable/) as well.
...
 
 - Open `http://127.0.0.1:5000/` in a browser. You will find in the presented UI 
-a link to the swagger.json with the definition of the api.
\ No newline at end of file
+a link to the swagger.json with the definition of the api.
+
+
+## CLI
+
+There are available commands aware of the API that can be verify useful for you. You
```
Angeluz-07: I think you meant 

> There are available commands that can be **very** useful for you.

cli.py
EliuX: By default it is actually localhost:5000 for development and localhost:8000 when it is using a wsgi server, like gunicorn. The idea behind setting a postman env variable is that you can set that env variable to whatever you want based on the environment you will be working at.
Nonetheless, I will create a param, so you can override the default param which by default will be localhost and everyone is happy :)

README.md
EliuX: LoL, thanks, I will fix it.

cli.py
EliuX: Yes, because I wanted to know whether the CLI file is actually running before any error inside it happens.

README.md
EliuX: LOL, yes, thanks for the catch.

cli.py
EliuX: Ok, sure.

README.md
EliuX: Yep

PR #24, opened 2020-03-13
Create configuration and database routing

EliuX:
This PR:

- Upgrades `pytest` to version `5.2.0`, so we can use `pytest-mock`.
- Tries out a mock in time_enties_namespace_test.py
- Add support for configurations. By default it works with `time_tracker_api.config.DefaultConfig`.
- Creates a configuration for a `whatever` database that can be taken as an example of what the access to any database should look like in a design/architecture that can be extensible. 
- whatever_repository.py how any database implementation (MongoDB, SQL databases, etc.) can be easily hooked using a generic Repository pattern.

In addition, I added a refactoring of the API models for what we thought would be a better naming system.


tests/time_entries/time_enties_namespace_test.py
```diff
@@ -0,0 +1,14 @@
+from flask import json
+
+def test_list_should_return_empty_array(mocker, client):
```
Angeluz-07: Could we have type hints here?
Also, check the name of the file, I think It should be 'time_entries_namespace_test' or why not just 'time_entries_test' ?

time_tracker_api/__init__.py
```diff
@@ -1,14 +1,37 @@
...
+               config_data=None):
     flask_app = Flask(__name__)
 
+    init_app_config(flask_app, config_path, config_data)
     init_app(flask_app)
 
     return flask_app
 
 
+def init_app_config(app, config_path, config_data=None):
```
Angeluz-07: Could we have type hints here.

time_tracker_api/database.py
```diff
@@ -0,0 +1,28 @@
...
+
+Put here your utils and class independent of
+the database solution
+"""
+from flask import Flask
+
+RepositoryModel = None
+
+
+def init_app(app: Flask) -> None:
```
Angeluz-07: Why not call this function 'init_database' directly ?

time_tracker_api/__init__.py
EliuX: Sure

time_tracker_api/database.py
EliuX: There is some sort of standard that some objects like the APIs in flask Restplus have a function called init_app so they take the Flask app as a parameter and they do whatever initialization logic they need to be ready to be used by that app. So whatever module or package you add if you want to make it like pluggable to Flask this is pretty much the way to go

tests/time_entries/time_enties_namespace_test.py
EliuX: Ok

Issue #26, opened 2020-03-17
Create examples for Flask Restplus models


EliuX:
We need to create examples of each template definition from the swagger specification.

PR #27, opened 2020-03-17
Add examples in flask - restplus models

Angeluz-07:
Examples have been added to make the swagger docs more expressive, with the help of the library [Faker](https://github.com/joke2k/faker).

Check #26  for details.

Also I included two fixes in code such as :
 - rename some variables called `uid` to `id` for consistency
 - rename a typo in a file name.


time_tracker_api/time_entries/time_entries_namespace.py
```diff
@@ -46,11 +54,13 @@
...
         title='Identifier',
         description='The unique identifier',
+        example=faker.uuid4()
     ),
     'running': fields.Boolean(
         readOnly=True,
         title='Is it running?',
-        description='Whether this time entry is currently running or not'
+        description='Whether this time entry is currently running or not',
+        example=faker.boolean()
```
EliuX: Actually, `running` has to be `true` if the `end_date` has no value. I think that `end_date` has to have a value or not.

PR #35, opened 2020-03-20
Creates repository sql database#25

EliuX:
This PR creates the base for using the SQL Repository in endpoint queries, so we are ready to persist data.


tests/resources.py
```diff
@@ -0,0 +1,13 @@
+from time_tracker_api.database import AuditedModel
+from time_tracker_api.sql_repository import db, SQLAuditedModel
+
+
+class PersonSQLModel(db.Model, SQLAuditedModel):
```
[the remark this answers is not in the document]
EliuX: Because this is the SQLAlchemy model, not a generic one

tests/sql_repository_test.py
```diff
@@ -0,0 +1,89 @@
...
+    """Updates an existing element"""
+    existing_element = existing_elements_registry[0]
+
+    updated_element = sql_repository.update(existing_element.id,
+                                            dict(name="Jon Snow", age=34))
+
+    assert updated_element is not None
+    assert updated_element.id == existing_element.id
+    assert updated_element.name == "Jon Snow"
+    assert updated_element.age == 34
```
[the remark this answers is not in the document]
EliuX: But I did not. The tests are not that complex and unless I use that value more than once it has no value for me. I would even have used a constant if it was in the codebase, but texts have to be easy to read and this makes perfect sense for me.

tests/resources.py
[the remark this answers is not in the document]
enriquezrene: is it extremelly necessat to include `SQL` as part of the name?

tests/sql_repository_test.py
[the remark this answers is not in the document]
enriquezrene: yes, you are using them twice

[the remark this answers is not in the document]
EliuX: You see it twice, but this is just only being used checking when I changed it before. You also see In another test an example alike where I also reference a part of this full name called `Snow`, and if I apply this logic of yours I would have to create a variable for that last name as well?

tests/resources.py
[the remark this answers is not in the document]
EliuX: Yes, because this model is specific to the SQL implementation. As well as I have the SQL prefix in the repository and other resources related to this particular technology, which has its own characteristics: specific library, handling of commits, rollbacks, the concept of tables, columns and so forth, which is no the same as, for instance, an In-memory implementation, a Mongo Implementation, etc.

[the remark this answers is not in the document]
EliuX: Just by looking it has the SQL prefix you know that this has SQL-specific logic.

tests/sql_repository_test.py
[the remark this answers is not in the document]
EliuX: In addition, I presented in Slack more evidence. But you can take as a reference these 2 videos related to tests from 2 great developers of different languages, where no variable for cases like this is used:
youtube.com/watch?v=LiyQ8bvLzIE
https://youtu.be/E87XhgYBM-Y?t=1684
In return, I would expect to see a good example to illustrate your case, so I can understand it better.
In any case, a nit-pick comment like this does not reason whatsoever for blocking this PR, so long and so important to solve many other tickets.

cli.py
```diff
@@ -44,6 +46,23 @@ def gen_postman_collection(filename='timetracker-api-postman-collection.json',
     save_data(parsed_json, filename)
 
 
+@cli_manager.command
```
enriquezrene: quick question: is there any chance to perform db/schema migrations by steps? 
If not, what we are going to do if we need to undo just the last change executed for example?

tests/conftest.py
```diff
@@ -1,18 +1,36 @@
...
 
 
 @pytest.fixture
 def client(app: Flask) -> FlaskClient:
     """A test client for the app."""
     with app.test_client() as c:
         return c
+
+
+@pytest.fixture(scope="module")
```
enriquezrene: What's the intention of this test?

tests/projects/projects_namespace_test.py
```diff
@@ -1,11 +1,17 @@
 from flask import json
+from flask.testing import FlaskClient
+from pytest_mock import MockFixture
 
 
-def test_list_should_return_nothing(client):
-    """Should return an empty array"""
+def test_list_all_elements(client: FlaskClient, mocker: MockFixture):
```
enriquezrene: No need to have comments here, you can change the method name as follows:

```
def test_endpoint_returns_all_elements_found_by_dao(client: FlaskClient, mocker: MockFixture):
    elements_found_by_dao = []
    from time_tracker_api.projects.projects_namespace import project_dao
    repository_find_all_mock = mocker.patch.object(project_dao.repository, 'find_all', return_value=elements_found_by_dao)

    response = client.get("/projects", follow_redirects=True)

    assert 200 == response.status_code
    
    json_data = json.loads(response.data)
    assert elements_found_by_dao == json_data
    repository_find_all_mock.assert_called_once()
```

cli.py
EliuX: This was not even requested in this issue. I just created these commands because of a need I had during my work with the database. 
I case something beyond a seeding is required we can tackle it down in another ticket. Flask migrates looks like a good choice at a glance.

tests/conftest.py
EliuX: This is not a test, these are fixtures, which are basically a factory where objects are created to be injected in the tests. It works basically as an IoC mechanism.

tests/resources.py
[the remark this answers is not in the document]
enriquezrene: We are not going to add support for Mongo, in-memory or so. We should not pollute code with implementation details.

tests/sql_repository_test.py
```diff
@@ -0,0 +1,89 @@
+from faker import Faker
+
+fake = Faker()
+
+existing_elements_registry = []
+last_deleted_element = None
+
+
+def test_create(sql_repository):
+    """Should create a new Entry"""
```
enriquezrene: why not naming the function/method name as 

`test_creates_a_new_entry`

Then you can get rid of comments

tests/sql_repository_test.py
```diff
@@ -0,0 +1,89 @@
...
+    assert result.created_by is not None
+    assert result.updated_at is None
+    assert result.updated_by is None
+
+    existing_elements_registry.append(result)
+
+
+def test_find(sql_repository):
+    """Should find created element"""
+    existing_element = existing_elements_registry[0]
```
enriquezrene: same here

tests/sql_repository_test.py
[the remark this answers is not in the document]
enriquezrene: I'd rather prefer to use my approach but it seems yours is ok as well.

tests/sql_repository_test.py
```diff
@@ -0,0 +1,89 @@
...
+
+    found_element = sql_repository.find(existing_element.id)
+
+    assert found_element is not None
+    assert found_element.id is not None
+
+
+def test_update(sql_repository):
+    """Updates an existing element"""
```
enriquezrene: no need to have comments, please remove them and make test name consistent

tests/projects/projects_namespace_test.py
EliuX: That is basically the JUni4 of writing code, which is pretty much discouraged. The JUnit 5 way of describing test functions with annotations is a good prove of it. The function name is just a unique container, but not a good way of providing a verbose description of the test.

tests/sql_repository_test.py
EliuX: This function just tests the function create of the repository, this is a simple and unique name. As I explained before the comment in the function is providing a verbose description of the test.

tests/sql_repository_test.py
EliuX: Same here, yes.

time_tracker_api/projects/projects_model.py
```diff
@@ -1,80 +1,86 @@
...
+            def __str___(self):
+                return "the project \"%s\"" % self.name
+
+        class ProjectSQLDao(SQLCRUDDao):
+            def __init__(self):
+                SQLCRUDDao.__init__(self, ProjectSQLModel)
+
+        return ProjectSQLDao()
+    else:
+        """
```
enriquezrene: I think we should get rid of all the in-memory implementation. 
1. We will need to implement this in-memory code for all DAOs
2. We are not going to need this in-memory solution
3. A Dao should has more implementations if we really need them. For now, just one sql implementation is good enough

tests/sql_repository_test.py
EliuX: In Java I would not have put comments, but this is Python. It is common here put comments even in modules. This is a test that tests the function called `update`. This is very simple.

time_tracker_api/projects/projects_namespace.py
```diff
@@ -19,15 +20,23 @@
...
+        required=False,
         title='Type',
-        max_length=30,
+        max_length=10,
         description='If it is `Costumer`, `Training` or other type',
-        example=faker.word(['Customer', 'Training']),
+        enum=PROJECT_TYPE.valid_type_values(),
+        example=faker.word(PROJECT_TYPE.valid_type_values()),
+    ),
+    'active': fields.Boolean(
```
enriquezrene: We can have a more flexible structure having a `status` field. This field can have values like:
- ACTIVE
- INACTIVE
- PAUSED, etc..

time_tracker_api/projects/projects_namespace.py
```diff
@@ -58,53 +69,42 @@ def get(self):
...
-
-    @ns.doc('put_project')
+    @ns.doc('update_project')
+    @ns.response(422, 'The data has an invalid format.')
+    @ns.response(409, 'This project already exists')
     @ns.expect(project_input)
     @ns.marshal_with(project)
     def put(self, id):
-        """Create or replace a project"""
+        """Updates a project"""
```
enriquezrene: No comment needed here or is this going to be used by swagger?

time_tracker_api/projects/projects_model.py
EliuX: This in-memory code will not be used, this is just basically for documentation purposes. I wanted the other people who work in the project to see how you can create other db mechanisms. And now that the Azure database serverless implementation is showing some issues, it is nice to be aware of how this work.

time_tracker_api/projects/projects_namespace.py
Angeluz-07: Yes, I found these comments here are used by swagger.

time_tracker_api/projects/projects_namespace.py
EliuX: You can put this in another ticket.

time_tracker_api/security.py
```diff
@@ -0,0 +1,32 @@
...
+    Generates a security key for development purposes
+    :return: str
+    """
+    global dev_secret_key
+    dev_secret_key = fake.password(length=16, special_chars=True, digits=True, upper_case=True, lower_case=True)
+    if app.config.get("FLASK_DEBUG", False):
+        print('*********************************************************')
+        print("The generated secret is \"%s\"" % dev_secret_key)
+        print('*********************************************************')
+    return dev_secret_key
```
enriquezrene: Again, no comments needed. You can express your intention using code

```
def generate_dev_secret_key():
    from time_tracker_api import flask_app as app
    global dev_secret_key = generate_secret_key()
    if app.config.get("FLASK_DEBUG", False):
        print('*********************************************************')
        print("The generated secret is \"%s\"" % dev_secret_key)
        print('*********************************************************')
    return dev_secret_key

generate_secret_key()
    return fake.password(length=16, special_chars=True, digits=True, upper_case=True, lower_case=True)
```

EliuX: This key is a flask requirement, it is no clear why this is required. I created this for developers who are going to use this code to know this.

cli.py
enriquezrene: gotcha.

enriquezrene: https://github.com/ioet/time-tracker-api/issues/36

tests/conftest.py
enriquezrene: great!!!

time_tracker_api/projects/projects_namespace.py
enriquezrene: cool

PR #40, opened 2020-03-24
Persist and test API namespace for projects

EliuX:
This PR:
- Fixes the issue related to the random lose of data
- Test persistence of the API namespace for projects

This will set the foundation for future tests of the persistence of API namespaces, covering 100% of it.

### DIFF
--- a/tests/projects/projects_namespace_test.py
+++ b/tests/projects/projects_namespace_test.py
@@ -106,7 +106,7 @@ def test_get_project_should_return_422_for_invalid_id_format(client: FlaskClient
     repository_find_mock.assert_called_once_with(str(invalid_id))
 
 
-def update_project_should_succeed_with_valid_data(client: FlaskClient, mocker: MockFixture):
+def test_update_project_should_succeed_with_valid_data(client: FlaskClient, mocker: MockFixture):
     from time_tracker_api.projects.projects_namespace import project_dao
 
     repository_update_mock = mocker.patch.object(project_dao.repository,
@@ -118,7 +118,7 @@ def update_project_should_succeed_with_valid_data(client: FlaskClient, mocker: M
 
     assert 200 == response.status_code
     fake_project == json.loads(response.data)
-    repository_update_mock.assert_called_once_with(valid_id, valid_project_data)
+    repository_update_mock.assert_called_once_with(str(valid_id), valid_project_data)
 
 
 def test_update_project_should_reject_bad_request(client: FlaskClient, mocker: MockFixture):
@@ -147,7 +147,9 @@ def test_update_project_should_return_not_found_with_invalid_id(client: FlaskCli
                                                  'update',
                                                  side_effect=NotFound)
 
-    response = client.put("/projects/%s" % invalid_id, json=valid_project_data, follow_redirects=True)
+    response = client.put("/projects/%s" % invalid_id,
+                          json=valid_project_data,
+                          follow_redirects=True)
 
     assert 404 == response.status_code
     repository_update_mock.assert_called_once_with(str(invalid_id), valid_project_data)
@@ -192,8 +194,8 @@ def test_delete_project_should_return_422_for_invalid_id_format(client: FlaskCli
     invalid_id = fake.company()
 
     repository_remove_mock = mocker.patch.object(project_dao.repository,
-                                               'remove',
-                                               side_effect=UnprocessableEntity)
+                                                 'remove',
+                                                 side_effect=UnprocessableEntity)
 
     response = client.delete("/projects/%s" % invalid_id, follow_redirects=True)
 
--- a/tests/projects/projects_namespace_test.py
+++ b/tests/projects/projects_namespace_test.py
@@ -1,16 +1,201 @@
+from faker import Faker
 from flask import json
 from flask.testing import FlaskClient
 from pytest_mock import MockFixture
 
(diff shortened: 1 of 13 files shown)


time_tracker_api/__init__.py
```diff
@@ -34,16 +35,18 @@ def init_app_config(app: Flask, config_path: str, config_data: dict = None):
...
-    from .database import init_app as init_database
+    from time_tracker_api.database import init_app as init_database
     init_database(app)
 
-    from .api import api
+    from time_tracker_api.api import api
     api.init_app(app)
 
     if app.config.get('DEBUG'):
+        app.logger.setLevel(logging.INFO)
```
enriquezrene: should not be `app.logger.setLevel(logging.DEBUG)`?

time_tracker_api/config.py
```diff
@@ -7,6 +7,7 @@ class Config:
     SECRET_KEY = generate_dev_secret_key()
     DATABASE_URI = os.environ.get('DATABASE_URI')
     PROPAGATE_EXCEPTIONS = True
+    RESTPLUS_VALIDATE = True
```
Angeluz-07: Good.

time_tracker_api/__init__.py
EliuX: `logging.INFO` is too detailed.

tests/projects/projects_namespace_test.py
```diff
@@ -1,16 +1,201 @@
...
+                                                 return_value=fake_project)
+
+    response = client.post("/projects", json=invalid_project_data, follow_redirects=True)
+
+    assert 400 == response.status_code
+    repository_create_mock.assert_not_called()
+
+
+def test_list_all_projects(client: FlaskClient, mocker: MockFixture):
+    from time_tracker_api.projects.projects_namespace import project_dao
```
Angeluz-07: Notice, you are calling `from time_tracker_api.projects.projects_namespace import project_dao` in all the tests.

tests/projects/projects_namespace_test.py
```diff
@@ -1,16 +1,201 @@
...
+                                                 'create',
+                                                 return_value=fake_project)
+
+    response = client.post("/projects", json=valid_project_data, follow_redirects=True)
+
+    assert 201 == response.status_code
+    repository_create_mock.assert_called_once_with(valid_project_data)
+
+
+def test_create_project_should_reject_bad_request(client: FlaskClient, mocker: MockFixture):
```
Angeluz-07: I think we can use the name `test_create_project_should_fail_with_invalid_request` for consistency with the previous request.

tests/projects/projects_namespace_test.py
```diff
@@ -1,16 +1,201 @@
...
+                                                 'create',
+                                                 return_value=fake_project)
+
+    response = client.post("/projects", json=invalid_project_data, follow_redirects=True)
+
+    assert 400 == response.status_code
+    repository_create_mock.assert_not_called()
+
+
+def test_list_all_projects(client: FlaskClient, mocker: MockFixture):
```
Angeluz-07: What about `test_get_all_projects` or `test_get_projects`?

tests/projects/projects_namespace_test.py
```diff
@@ -1,16 +1,201 @@
...
+                                               'find',
+                                               side_effect=UnprocessableEntity)
+
+    response = client.get("/projects/%s" % invalid_id, follow_redirects=True)
+
+    assert 422 == response.status_code
+    repository_find_mock.assert_called_once_with(str(invalid_id))
+
+
+def update_project_should_succeed_with_valid_data(client: FlaskClient, mocker: MockFixture):
```
Angeluz-07: You are missing the `test_` prefix.

tests/projects/projects_namespace_test.py
```diff
@@ -1,16 +1,201 @@
...
+
+    valid_id = fake.random_int(1, 9999)
+    response = client.put("/projects/%s" % valid_id, json=valid_project_data, follow_redirects=True)
+
+    assert 200 == response.status_code
+    fake_project == json.loads(response.data)
+    repository_update_mock.assert_called_once_with(valid_id, valid_project_data)
+
+
+def test_update_project_should_reject_bad_request(client: FlaskClient, mocker: MockFixture):
```
Angeluz-07: I think we can use the name `test_update_project_should_fail_with_invalid_request` as before.

tests/projects/projects_namespace_test.py
EliuX: Yes, if I do it before the project_dao would be broken because it has to be constructed after the app instance is ready. This is a necessary evil.

tests/projects/projects_namespace_test.py
EliuX: I would rather to leave bad request because that is the verbose name of the HTTP status code `400`.

tests/projects/projects_namespace_test.py
EliuX: The description in the comments says list. As long as it makes sense and it is not ambiguous we are Ok.

tests/projects/projects_namespace_test.py
EliuX: Nice catch

tests/projects/projects_namespace_test.py
EliuX: Bad request is better

Issue #13, opened 2020-03-11
First implementation of API of time entries


EliuX:
We have defined the API for time-entries, but it is hollow, we must make it work. To do so let´s work in the biz-logic of its endpoints and its model.

PR #43, opened 2020-03-25
Implementation and test of api namespaces for time entries #13

EliuX:
This PR:
- Creates the implementation for the API namespace for time entries
- Creates the tests for the API namespace for time entries
- Replaces the status codes for constants
- Improves documentation of expected status codes for the API namespace for projects and time entries

### DIFF
--- a/time_tracker_api/api.py
+++ b/time_tracker_api/api.py
@@ -4,8 +4,7 @@ import sqlalchemy
 from faker import Faker
 from flask import current_app as app
 from flask_restplus import Api, fields
-from flask_restplus.fields import Raw, to_marshallable_type, MarshallingError
-from werkzeug.exceptions import Conflict
+from flask_restplus._http import HTTPStatus
 
 faker = Faker()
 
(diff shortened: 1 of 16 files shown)


time_tracker_api/api.py
```diff
@@ -4,6 +4,9 @@
 from faker import Faker
 from flask import current_app as app
 from flask_restplus import Api, fields
+from flask_restplus._http import HTTPStatus
+from flask_restplus.fields import Raw, to_marshallable_type, MarshallingError
+from werkzeug.exceptions import Conflict
```
Angeluz-07: I don't see how the two last imports are being used.

EliuX: Sure,  I will remove them

time_tracker_api/sql_repository.py
```diff
@@ -27,20 +29,15 @@ def init_app(app: Flask) -> None:
     db = SQLAlchemy(app)
 
     global SQLModel
-
-    class SQLModelClass(DatabaseModel):
-        def to_dto(self):
-            return self
-
-    SQLModel = SQLModelClass
+    SQLModel = DatabaseModel
```
Angeluz-07: Checking at changes of `DatabaseModel` Class, and tracing use of `SQLModel` It seems to me that these two classes are not useful anymore.

time_tracker_api/database.py
```diff
@@ -9,60 +9,47 @@
...
     def fresh(self):
-        raise NotImplementedError
+        raise NotImplementedError  # pragma: no cover
 
     def __call__(self, *args, **kwargs):
-        self.run()
+        self.run()  # pragma: no cover
 
 
 class DatabaseModel:
```
Angeluz-07: As I said in another comment. https://github.com/ioet/time-tracker-api/pull/43/files#r398215021 I think this class is not needed anymore.

time_tracker_api/sql_repository.py
EliuX: Whatever particularity that has to be inherited by the SQLModels should be done throw the SQL model. The `DatabaseModel` is platform-agnostic. Before we had this `to_dto` function that was required by such models. This is basically an encapsulation.

time_tracker_api/database.py
EliuX: Ok, I will remove then ;)

EliuX: Done

PR #44, opened 2020-03-25
Implement activities model

Angeluz-07:
This PR includes :
 - Implementation for the Activity Model
 - Suite of test for the Activity Model

I reused most of the implementation of projects. Thanks @EliuX for your help on it.

### DIFF
--- a/time_tracker_api/activities/activities_namespace.py
+++ b/time_tracker_api/activities/activities_namespace.py
@@ -1,7 +1,7 @@
 from faker import Faker
 from flask_restplus import fields, Resource, Namespace
+from flask_restplus._http import HTTPStatus
 
-from time_tracker_api import flask_app
 from time_tracker_api.api import audit_fields
 from time_tracker_api.activities.activities_model import create_dao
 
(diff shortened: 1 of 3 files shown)


time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
 
-activities_dao = create_dao()
+activity_dao = create_dao()
 
 
 @ns.route('')
 class Activities(Resource):
     @ns.doc('list_activities')
-    @ns.marshal_list_with(activity, code=200)
+    @ns.marshal_list_with(activity, code=HTTPStatus.OK)
```
EliuX: by default it is `OK`. So this is redundant.

time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
     def get(self):
-        return []
+        """List all activities"""
+        return activity_dao.get_all(), HTTPStatus.OK
 
     @ns.doc('create_activity')
+    @ns.response(HTTPStatus.BAD_REQUEST, 'Bad request')
```
EliuX: If we are going to put a message, lets better put one customized, like:
> Invalid format or structure of the attributes of the activity

Also, it would be nice to specify that a CONFLICT may occur if there is already an activity with the same name
```
    @ns.response(HTTPStatus.CONFLICT, 'This activity already exists')
```

time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
+        """Create an activity"""
+        return activity_dao.create(ns.payload), HTTPStatus.CREATED
 
 
 @ns.route('/<string:id>')
-@ns.response(404, 'Activity not found')
-@ns.param('id', 'The unique identifier of the activity')
+@ns.response(HTTPStatus.NOT_FOUND, 'Activity not found')
+@ns.param('id', 'The activity identifier')
 class Activity(Resource):
```
EliuX: Everytime the id is not well specified, like with an invalid format, lets say an string `asd`, you will get 
a 422 error. So lets specify
```
@ns.response(HTTPStatus.UNPROCESSABLE_ENTITY, 'The id has an invalid format')
```

time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
     @ns.doc('get_activity')
-    @ns.marshal_with(activity)
+    @ns.response(HTTPStatus.UNPROCESSABLE_ENTITY, 'The id has an invalid format')
```
EliuX: Instead of here, or you will have to repeat it every time.

time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
+    @ns.marshal_with(activity, code=HTTPStatus.OK)
     def get(self, id):
-        return {}
-
-    @ns.doc('delete_activity')
-    @ns.response(204, 'The activity was deleted successfully (No content is returned)')
-    def delete(self, id):
-        return None, 204
+        """Get an activity"""
+        return activity_dao.get(id), HTTPStatus.OK
```
EliuX: Not necessary to specify OK as a result. By default it returns 200.

time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
 
-    @ns.doc('put_activity')
-    @ns.response(400, 'Invalid format of the attributes of the activity.')
+    @ns.doc('update_activity')
+    @ns.response(HTTPStatus.UNPROCESSABLE_ENTITY, 'The data has an invalid format.')
```
EliuX: I would suggest to remove `UNPROCESSABLE_ENTITY` (it would be in the class) and add:
```
    @ns.response(HTTPStatus.BAD_REQUEST, 'Invalid format or structure of the activity')
    @ns.response(HTTPStatus.CONFLICT, 'An activity already exists with this new data')
```

time_tracker_api/activities/activities_namespace.py
```diff
@@ -41,41 +43,49 @@
...
     @ns.expect(activity_input)
     @ns.marshal_with(activity)
     def put(self, id):
-        return ns.payload
+        """Update an activity"""
+        return activity_dao.update(id, ns.payload)
+
+    @ns.doc('delete_activity')
+    @ns.response(HTTPStatus.UNPROCESSABLE_ENTITY, 'The id has an invalid format')
```
EliuX: Remove the `UNPROCESSABLE_ENTITY`. It will be in the class.

PR #68, opened 2020-04-06
Add project type and customer models

### LINKED ISSUE #67: Create ProjectType and Customer Models with current implementation
Create these models using current implementation. This is required because UI team needs an API with these two models to work with.

Angeluz-07:
Fixes #67

### DIFF
--- a/time_tracker_api/project_types/project_types_model.py
+++ b/time_tracker_api/project_types/project_types_model.py
@@ -1,0 +1,35 @@
+from time_tracker_api.database import CRUDDao
+
+
+class ProjectTypeDao(CRUDDao):
+    pass
+
+
+def create_dao() -> ProjectTypeDao:
+    from time_tracker_api.sql_repository import db
+    from time_tracker_api.database import COMMENTS_MAX_LENGTH
+    from time_tracker_api.sql_repository import SQLCRUDDao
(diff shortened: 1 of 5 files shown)


time_tracker_api/project_types/project_types_model.py
```diff
@@ -0,0 +1,35 @@
...
+    import uuid
+
+    class ProjectTypeSQLModel(db.Model):
+        __tablename__ = 'project_type'
+        id = db.Column(UUIDType(binary=False), primary_key=True, default=uuid.uuid4)
+        name = db.Column(db.String(50), unique=True, nullable=False)
+        description = db.Column(db.String(COMMENTS_MAX_LENGTH), unique=False, nullable=False)
+        parent_id = db.Column(UUIDType(binary=False), default=uuid.uuid4)
+        customer_id = db.Column(UUIDType(binary=False), default=uuid.uuid4)
+        deleted = db.Column(UUIDType(binary=False), default=uuid.uuid4)
```
EliuX, line 22: This and the other SQL models should inherit from `AuditedSQLModel`. `AuditedSQLModel` should also contain the common fields, as `deleted` and if possible add to it an auto-calculated attribute called `visible` which is True if the next condition is:
```
self.deleted == None
```

PR #74, opened 2020-04-09
Closes #36 Allow to run migrations

### LINKED ISSUE #36: add database/schema migrations support
The API works using a SQL database, we need to add support to DB/schema migrations so we can make evolve the DB structure whenever it is needed.

**Acceptance Criteria**
- DB/schema migrations support is added (https://en.wikipedia.org/wiki/Schema_migration)
- README file should be updated with required instructions to run db migrations up and down
- Current db seeder implementation should be revisited and removed

EliuX:
In this PR it is installed and configured a library to run migrations. In addition, it creates an initial migration that creates an initial database schema. With containers for:
- Project
- Project type
- Activity
- Customer
- Time entry

Although the library for migrations we are using [is not perfect](https://github.com/Lieturd/migrate-anything/issues/3), we were able to run the migrations and persist them in Azure Cosmos DB.

### DIFF
--- a/time_tracker_api/project_types/project_types_model.py
+++ b/time_tracker_api/project_types/project_types_model.py
@@ -1,3 +1,5 @@
+from azure.cosmos import PartitionKey
+
 from time_tracker_api.database import CRUDDao
 
 
@@ -33,3 +35,15 @@ def create_dao() -> ProjectTypeDao:
             SQLCRUDDao.__init__(self, ProjectTypeSQLModel)
 
     return ProjectTypeSQLDao()
+
+
+container_definition = {
+    'id': 'project_type',
+    'partition_key': PartitionKey(path='/customer_id'),
+    'unique_key_policy': {
+        'uniqueKeys': [
+            {'paths': ['/name']},
+            {'paths': ['/deleted']},
+        ]
+    }
+}
--- a/time_tracker_api/projects/projects_model.py
+++ b/time_tracker_api/projects/projects_model.py
@@ -1,5 +1,7 @@
 import enum
 
+from azure.cosmos import PartitionKey
+
 from time_tracker_api.database import CRUDDao
 
 
@@ -44,3 +46,15 @@ def create_dao() -> ProjectDao:
             SQLCRUDDao.__init__(self, ProjectSQLModel)
 
     return ProjectSQLDao()
+
+
+container_definition = {
+    'id': 'project',
+    'partition_key': PartitionKey(path='/tenant_id'),
+    'unique_key_policy': {
+        'uniqueKeys': [
+            {'paths': ['/name']},
+            {'paths': ['/deleted']},
+        ]
+    }
+}
--- a/time_tracker_api/time_entries/time_entries_model.py
+++ b/time_tracker_api/time_entries/time_entries_model.py
@@ -1,3 +1,4 @@
+from azure.cosmos import PartitionKey
 from sqlalchemy_utils import ScalarListType
 
 from time_tracker_api.database import CRUDDao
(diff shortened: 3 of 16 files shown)


time_tracker_api/project_types/project_types_model.py
```diff
@@ -33,3 +35,15 @@ def __init__(self):
             SQLCRUDDao.__init__(self, ProjectTypeSQLModel)
 
     return ProjectTypeSQLDao()
+
+
+container_definition = {
+    'id': 'project_type',
+    'partition_key': PartitionKey(path='/customer_id'),
+    'unique_key_policy': {
+        'uniqueKeys': [
```
Angeluz-07, line 44: I think we need to include parent_id and maybe also tenant_id.

time_tracker_api/projects/projects_model.py
```diff
@@ -44,3 +46,15 @@ def __init__(self):
             SQLCRUDDao.__init__(self, ProjectSQLModel)
 
     return ProjectSQLDao()
+
+
+container_definition = {
+    'id': 'project',
+    'partition_key': PartitionKey(path='/tenant_id'),
+    'unique_key_policy': {
```
Angeluz-07, line 54: I think we need project_type and maybe customer_id

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -47,3 +48,15 @@ def __init__(self):
...
 
     return TimeEntriesSQLDao()
+
+
+container_definition = {
+    'id': 'time_entry',
+    'partition_key': PartitionKey(path='/tenant_id'),
+    'unique_key_policy': {
+        'uniqueKeys': [
+            {'paths': ['/owner_id', '/end_date']},
```
Angeluz-07, line 58: These two fields belongs to the model. Why is this path different form the others to define uniqueKeys. ?

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -47,3 +48,15 @@ def __init__(self):
             SQLCRUDDao.__init__(self, TimeEntrySQLModel)
 
     return TimeEntriesSQLDao()
+
+
+container_definition = {
+    'id': 'time_entry',
+    'partition_key': PartitionKey(path='/tenant_id'),
+    'unique_key_policy': {
+        'uniqueKeys': [
```
Angeluz-07, line 57: Following the same observation of models above. I think here we need project_id, activity_id.

time_tracker_api/projects/projects_model.py
EliuX: No, you can actually have multiple projects with the same type in a same tenant. That is not a problem.

time_tracker_api/time_entries/time_entries_model.py
EliuX: Because here the uniqueness is the combination `owner_id` and `end_date`. This means that there cannot be 2 entries with the same `end_date`. Having into consideration that `null` is also a value, then if by mistake the client sends the creation of a time entry twice, as the second one will also have the end_date as null there will be a uniqueness issue and it will not be created. This way the client will be forced to close the existent opened time entry to create a new one.

time_tracker_api/time_entries/time_entries_model.py
EliuX: No, you can be working in that same project_id and doing that same activity id even with different technologies or even the same combination of the 3 of them. For instance, you are working at something and then you change to the standup and when you finish you go back to a time entry with the same data as before.

time_tracker_api/project_types/project_types_model.py
EliuX: The tenant id is the parent partition_key, so it kind of defines a domain for those unique keys (it is kind of unique by default). The `parent_id` cannot be unique, because this is a tree structure: every branch has multiple branches or leaves not only one.

PR #76, opened 2020-04-16
Create models in Cosmos DB SQL API for new DAL 

EliuX:
This PR creates the models and the correspondent tests (for a 95% of coverage) for the API models of:
- projects
- activities
- project types
- time entries
- customers

It also checks whether a time entry collides with other items in the database, based on its date range. This keeps a consistent set of time entries
It also allows specifying an array in the repository with the fields that will be used to order the list of items.


time_tracker_api/activities/activities_model.py
```diff
@@ -1,46 +1,50 @@
...
+@dataclass()
+class ActivityCosmosDBModel(CosmosDBModel):
+    id: str
+    name: str
+    description: str
+    deleted: str
+    tenant_id: str
+
+    def __init__(self, data):
+        super(ActivityCosmosDBModel, self).__init__(data)  # pragma: no cover
```
Angeluz-07, line 33: I think here It should CosmosDBModel instead.

PR #83, opened 2020-04-17
fix: Close #82 Enforce users manipulate only their time entries

### LINKED ISSUE #82: Users should only access time entries that belongs to them
In all actions of the API namespace for time entries, we should make sure that the data that is going to be manipulated actually belongs to the requesting user.

TODO:
Apply the necessary changes to the DAO of the time entries and its corresponding tests.

EliuX:
In this PR we make possible change the model of the time entries, so that it is checked that the current user can only manipulate time entry items where he is the owner, otherwise he will get an 403 error:
image

This is possible because of the new parameter `peeker` and `conditions`:
* peeker: It is a peeker that is meant to watch the item that is manipulated and if a biz-logic criteria requirement is not fulfilled it can short-circuit the user request flow by raising an `CustomError`.
* conditions: It is a python dictionary that contains the attribute and the corresponding expected value we expect the entries have in the database. This is done by creating an equality condition in the where clause of the SQL query executed in the function `find_all`.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -9,6 +9,7 @@ export FLASK_APP=time_tracker_api
 ## For Azure Cosmos DB
 export DATABASE_ACCOUNT_URI=https://<project_db_name>.documents.azure.com:443
 export DATABASE_MASTER_KEY=<db_master_key>
-export DATABASE_NAME=<db_name>
 ### or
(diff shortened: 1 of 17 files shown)


Angeluz-07:
Good Job.

PR #90, opened 2020-04-20
Serve swagger.json on UI over https

### LINKED ISSUE #47: Https is not working for main page
We have problems accessing the main page by HTTPS, but through HTTP everything works correctly. All endpoints are available via HTTPS or HTTP the problem is the main page. The API does not provide logs about this problem.
HTTPS: https://timetracker-api.azurewebsites.net/
HTTP: http://timetracker-api.azurewebsites.net/

Angeluz-07:
Fixes #47

### DIFF
--- a/time_tracker_api/__init__.py
+++ b/time_tracker_api/__init__.py
@@ -46,6 +46,8 @@ def init_app(app: Flask):
         app.logger.setLevel(logging.INFO)
         add_debug_toolbar(app)
 
+    add_werkzeug_proxy_fix(app)
+
     cors_origins = app.config.get('CORS_ORIGINS')
     if cors_origins:
         enable_cors(app, cors_origins)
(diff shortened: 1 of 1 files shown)


time_tracker_api/__init__.py
```diff
@@ -74,3 +76,9 @@ def enable_cors(app: Flask, cors_origins: str):
     cors_origins_list = cors_origins.split(",")
     CORS(app, resources={r"/*": {"origins": cors_origins_list}})
     app.logger.info("Set CORS access to [%s]" % cors_origins)
+
+
+def add_proxy_fix_to_serve_swagger_over_https(app: Flask):
+    from werkzeug.middleware.proxy_fix import ProxyFix
```
EliuX, line 82: I would call this function `add_fixers` and in the commit then I would say 
> Add ProxyFix to serve swagger over HTTPs.

This is supposed to be registered in the CHANGELOG and provide this node.
You can leave the logged message because it serves to explain what this particular fix is for.

Angeluz-07: I agree with the commit message change. But I don't agree with the function name "add_fixers". It seems to general for me. This is a fix for a very specific problem. If later we found we need more fixers, it'll be ok to change the name to "add_fixers". But for now I think we are ok leaving the name of the function specific.

EliuX: Ok, but that function explains reasons and not what that actual function does. If you want to explain the reasons in a verbose way, better use a logger. But I understand your point on trying to be more specific, so I would name it `use_werkzeug_proxy_fix`.

Issue #91, opened 2020-04-20
Find all items filtering by attributes


EliuX:
Right now every API namespace has a `find all` endpoint, which basically returns a list of all available items. We would like to be able to filter that result by attributes of interest. E.g.:
- name
- deleted

This way we could answer questions like:
> Are there any deleted item having "x" as name 

TODO:
* Implement a way that allows finding with any repository all items that match a particular criteria: 
   - The attribute is equal to x
  - The attribute partially matches x (Optionally, if Cosmos DB SQL API allows it)
  - The attribute is not equal to (Nice to have)
* Create the corresponding search criteria for every API namespace. It should be available in a URL like `GET https://timetracker-api.azurewebsites.net/projects?filters[name]=ernest%20and%20young&filters[deleted]=true`

PR #95, opened 2020-04-21
fix: #91 adding filter by attributes

enriquezrene:
fix: #91 adding filter by attributes

### DIFF
--- a/tests/time_tracker_api/activities/activities_namespace_test.py
+++ b/tests/time_tracker_api/activities/activities_namespace_test.py
@@ -67,7 +67,8 @@ def test_list_all_activities(client: FlaskClient,
     assert HTTPStatus.OK == response.status_code
     json_data = json.loads(response.data)
     assert [] == json_data
-    repository_find_all_mock.assert_called_once_with(partition_key_value=tenant_id)
+    repository_find_all_mock.assert_called_once_with(partition_key_value=tenant_id,
+                                                     conditions={})
 
 
 def test_get_activity_should_succeed_with_valid_id(client: FlaskClient,
--- a/commons/data_access_layer/cosmos_db.py
+++ b/commons/data_access_layer/cosmos_db.py
@@ -176,7 +176,8 @@ class CosmosDBRepository:
 
     def partial_update(self, id: str, changes: dict, partition_key_value: str,
                        peeker: 'function' = None, visible_only=True, mapper: Callable = None):
-        item_data = self.find(id, partition_key_value, peeker=peeker, visible_only=visible_only, mapper=dict)
+        item_data = self.find(id, partition_key_value, peeker=peeker,
+                              visible_only=visible_only, mapper=dict)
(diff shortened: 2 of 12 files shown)


tests/time_tracker_api/activities/activities_namespace_test.py
```diff
@@ -57,6 +58,21 @@ def test_list_all_activities(client: FlaskClient, mocker: MockFixture):
     repository_find_all_mock.assert_called_once()
 
 
+def test_list_all_activities_with_conditions(client: FlaskClient, mocker: MockFixture):
+    from time_tracker_api.activities.activities_namespace import activity_dao
+    repository_find_all_mock = mocker.patch.object(activity_dao.repository,
+                                                   'find_all',
+                                                   return_value=[])
+
+    response = client.get("/activities?a=b", follow_redirects=True)
```
EliuX, line 67: This is not taking into account that the attributes of the filter are not the unique URL params that are passed. Other params can be, for instance, order attributes and pagination variables (offset, pageSize). That is why in #91 we proposed to enclose those filter attributes in an array, like `filter`.

time_tracker_api/customers/customers_namespace.py
```diff
@@ -45,7 +46,7 @@ class Customers(Resource):
     @ns.marshal_list_with(customer)
     def get(self):
         """List all customers"""
-        return customer_dao.get_all()
+        return customer_dao.get_all(conditions=request.args)
```
EliuX, line 49: What if you have some query like this?
```
/activities?a=1&b=2&pageSize=0&orders["start_date"]="DESC"&limit=1
```
This is not going to work

EliuX: In addition, it would be nice to filter per namespace that the attributes you want to filter by are valid.

enriquezrene: Good catch

enriquezrene: > What if you have some query like this?
> 
> ```
> /activities?a=1&b=2&pageSize=0&orders["start_date"]="DESC"&limit=1
> ```
> 
> This is not going to work, because `a` and `b` are the attributes but you are filtering for whatever it is specified in the request arguments.

Do we have filtering and sorting arguments right now?

EliuX:
Also, please rename the commit message.
Instead of `fix`, which is a patch, use `feat`, because you are adding a new feature
For Github you can use Fix or Close and the number of the issue, so, an example final message for the commit could be:
>feat: Close #91 Add filter by attributes to get all

time_tracker_api/customers/customers_namespace.py
EliuX: The repository has sorting capabilities right now. For instance, The repository for time entries orders the results by `start_date`. We could do this dynamically in the future.

PR #99, opened 2020-04-22
Feature/enforce commit message style#78

### LINKED ISSUE #78: Enforce semantic versioning
It seems that if we create a commit with a semantic version, the CI creates a new release. But as humans, as we are, we might forget to do such a thing. So in a pre-commit-hook we should enforce the usage of the semantic version, so we can get our versions increased in the API version. Any other action or configuration to make this possible will be welcome.

Angeluz-07:
Fixes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,9 @@ automatically [pip](https://pip.pypa.io/en/stable/) as well.
     
     
     Remember to do it with Python 3.
-    
+
+- Run `pre-commit install`. For more details, check out Development > Git hooks.
+
 ### How to use it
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -74,6 +74,16 @@ DateTime strings in Azure Cosmos DB is `YYYY-MM-DDThh:mm:ss.fffffffZ` which foll
 
 ## Development
 
+### Git hooks
```
EliuX, line 77: Let's reference this section as a step in the **Setup**.
* Run `pre-commit install`. For more details, check out _Development > Git hooks_.

PR #115, opened 2020-04-27
fix: #114 description should be optional

enriquezrene:
fix: #114 description should be optional on time-entry creation


time_tracker_api/time_entries/time_entries_namespace.py
```diff
@@ -35,7 +35,7 @@
         description='The id of the selected activity',
         example=faker.uuid4(),
     ),
-    'description': fields.String(
```
EliuX, line 38: If you change this description better change it in the other namespaces.

enriquezrene: makes sense, I will

PR #117, opened 2020-04-28
Feature/get time entries per month and year#106

### LINKED ISSUE #106: Get time-entries per month of the year
We need an endpoint able to return all the time entries in a specified month for my user. The endpoint should look like:

GET /time-entries

The user id read from the JWT. This endpoint should also allow filtering by dates as follows:

GET /time-entries?month={MONTH}&year={YEAR}

Where month can be:
1 = January
2 = February, etc

If YEAR is not specified, then the backend will return all the records with a start time matching the specified month in the current year. If MONTH and YEAR are not specified, then all the time-entries starting in the current month and year should be returned.

Angeluz-07:
Fixes #106


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -74,6 +74,35 @@ def create_sql_ignore_id_condition(id: str):
...
+    def create_sql_date_range_filter(custom_args: dict) -> str:
+        if 'start_date' and 'end_date' in custom_args:
+            return """
+            ((c.start_date BETWEEN @start_date AND @end_date) OR
+             (c.end_date BETWEEN @start_date AND @end_date))
+            """
+        else:
+            return ''
+
+    def find_all(self, partition_key_value: str, conditions: dict, custom_args: dict):
```
EliuX, line 87: The find_all in time entries should be more personalized. This function can and should be aware of parameters related to the month and year you want to query (you can create 2 params or better one for this kind of filtering). That will be better than `custom_args`, which is ambiguous for this purpose.

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -74,6 +74,35 @@ def create_sql_ignore_id_condition(id: str):
...
+        custom_conditions = []
+        custom_conditions.append(
+            self.create_sql_date_range_filter(custom_args)
+        )
+
+        custom_params = [
```
EliuX, line 93: You should have a function that knows the params logic for this particular case. So this function will create the list of dictionary entries you need to add for the query to run. That logic should not be inside of this function `find_all`.

time_tracker_api/time_entries/time_entries_namespace.py
```diff
@@ -108,10 +108,20 @@
     "uri",
 ])
 
+# custom attributes filter
+attributes_filter.add_argument('month', required=False,
+                                    store_missing=False,
+                                    help="(Filter) month to filter",
```
EliuX, line 114: LoL, lets better the filter prefix, which I added to the description of fields to make it clearer that it is a description of a filter. Does that make sense?
I would leave a help message like:
>Month to filter by

time_tracker_api/time_entries/time_entries_namespace.py
```diff
@@ -108,10 +108,20 @@
...
+                                    location='args')
+attributes_filter.add_argument('year', required=False,
+                                    store_missing=False,
+                                    help="(Filter) year to filter",
```
EliuX, line 118: >Year to filter by

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -170,8 +200,29 @@ def check_whether_current_user_owns_item(cls, data: dict):
 
     def get_all(self, conditions: dict = {}) -> list:
         conditions.update({"owner_id": self.current_user_id()})
+
+        if 'month' and 'year' in conditions:
```
EliuX, line 204: Too much logic for a single function don't you think? Clean code my friend.

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -170,8 +200,29 @@ def check_whether_current_user_owns_item(cls, data: dict):
...
+            month = int(conditions.get("month"))
+            year = int(conditions.get("year"))
+            conditions.pop('month')
+            conditions.pop('year')
+        elif 'month' in conditions:
+            month = int(conditions.get("month"))
+            year = get_current_year()
+            conditions.pop('month')
+        else:
+            month = get_current_month()
```
EliuX, line 214: The thing is that if they are not present in the filter you should actually skip that filter. It has better performance and better consistency: When a filter is not present it is because it is not needed. For less than this, we avoided having into account whether the deleted field was defined... and also `null`.

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -170,8 +200,29 @@ def check_whether_current_user_owns_item(cls, data: dict):
...
+            year = get_current_year()
+
+        start_date, end_date = get_date_range_of_month(year, month)
+
+        custom_args = {
+            'start_date': start_date.isoformat(),
+            'end_date': end_date.isoformat()
```
EliuX, line 221: Add trailing a comma

time_tracker_api/time_entries/time_entries_model.py
EliuX: This part may return an empty dictionary or whatever default value expects `find_all` for time entries for this parameter.

PR #131, opened 2020-05-11
Don't display projects that belong to deleted customers

fabidick22:
Ticket: https://github.com/orgs/ioet/projects/5#card-37540366

### DIFF
--- a/time_tracker_api/projects/projects_model.py
+++ b/time_tracker_api/projects/projects_model.py
@@ -68,15 +68,11 @@ class ProjectCosmosDBDao(APICosmosDBDao, ProjectDao):
         event_ctx = self.create_event_context("read-many")
         customer_dao = customers_create_dao()
         customers = customer_dao.get_all()
-        projects = self.repository.find_all(event_ctx, **kwargs)
-        active_projects = []
 
-        for project in projects:
-            for customer in customers:
-                if customer in project:
-                    active_projects.append(project)
-                    break
-        return active_projects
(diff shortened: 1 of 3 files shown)


time_tracker_api/projects/projects_model.py
```diff
@@ -34,19 +34,52 @@ class ProjectCosmosDBModel(CosmosDBModel):
...
+        """
+        Get all the projects an active client has
+        :param (dict) conditions: Conditions for querying the database
+        :param (dict) kwargs: Pass arguments
+        :return (list): ProjectCosmosDBModel object list
+        """
+        event_ctx = self.create_event_context("read-many")
+        customer_dao = customers_create_dao()
+        customers = customer_dao.get_all()
+        projects = self.repository.find_all(event_ctx, **kwargs)
```
enriquezrene, line 71: I think you should change this a little bit. 
1. Query all customers (DONE and OK)
2. Query projects `where customer_id IN ( customer_id from the above-query executed)`

With the IN constraint, you will have only projects that belong to customers that are not deleted.

fabidick22: So you say, do a query for each customer to find the projects?
I think it is a good solution, but we have to keep in mind the response time for each query, if we have many customers, this solution may be a little slow.

enriquezrene: Not at all, one query with an IN clause to get all the projects

fabidick22: Got it

PR #134, opened 2020-05-12
Feat(time-entries): include project name in time-entries response

### LINKED ISSUE #129: Include project name on time-entries response
We want to display the project name when time-entries are show in the UI so we need to have this value included as part of the response of GET /time-entries

Angeluz-07:
Close #129

### DIFF
--- a/time_tracker_api/time_entries/time_entries_model.py
+++ b/time_tracker_api/time_entries/time_entries_model.py
@@ -1,6 +1,7 @@
 import abc
 from dataclasses import dataclass, field
 from typing import List, Callable
+from flask import jsonify
 
 from azure.cosmos import PartitionKey
 from flask_restplus._http import HTTPStatus
@@ -18,6 +19,10 @@ from commons.data_access_layer.cosmos_db import (
 from commons.data_access_layer.database import EventContext
 
 from time_tracker_api.time_entries.custom_modules import worked_time
+from time_tracker_api.time_entries.custom_modules.utils import (
+    add_project_name_to_time_entries,
+)
+from time_tracker_api.projects import projects_model
 from time_tracker_api.database import CRUDDao, APICosmosDBDao
 from time_tracker_api.security import current_user_id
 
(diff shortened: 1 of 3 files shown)


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -106,6 +111,24 @@ def create_sql_date_range_filter(date_range: dict) -> str:
         else:
             return ''
 
+    def find(
+        self,
+        id: str,
+        event_context: EventContext,
+        peeker: 'function' = None,
```
fabidick22, line 118: `'function'` what kind of data is this?

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -119,14 +142,18 @@ def find_all(
 
         custom_params = self.generate_params(date_range)
 
-        return CosmosDBRepository.find_all(
+        time_entries = CosmosDBRepository.find_all(
```
enriquezrene, line 145: is this returning the time-entries belonging to the JWT user or all of them?
Can you please restrict to only the time-entries belonging to the user in the JWT?

time_tracker_api/time_entries/time_entries_model.py
Angeluz-07: Is a type hinting supposed to represent that this parameter receives a function. Is part of a previous implementation. Looking on internet, It seems the right type hint is callable -> https://stackoverflow.com/questions/45897225/what-is-the-type-hint-for-a-function I guess the current notation was preferred over Callable, because it looks more expressive about the type of the parameter.

time_tracker_api/time_entries/time_entries_model.py
Angeluz-07: >is this returning the time-entries belonging to the JWT user or all of them?

Yes. That is how it works.

Check out the user_id/owner_id value is passed in the dictionary conditions, in the dao:
```
class TimeEntriesCosmosDBDao(APICosmosDBDao, TimeEntriesDao):
...
   def get_all(self, conditions: dict = {}) -> list:
        event_ctx = self.create_event_context("read-many")
        conditions.update({"owner_id": event_ctx.user_id})

        date_range = self.handle_date_filter_args(args=conditions)
        return self.repository.find_all(
            event_ctx, conditions=conditions, date_range=date_range
        )
```

time_tracker_api/time_entries/time_entries_model.py
Angeluz-07: So it's ok to change this for a better design.  I opened a ticket for it https://github.com/ioet/time-tracker-backend/issues/135

PR #143, opened 2020-05-18
Fix/return all time entries range time#122

fabidick22:


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -102,6 +111,7 @@ def create_sql_ignore_id_condition(id: str):
 
     @staticmethod
     def create_sql_date_range_filter(date_range: dict) -> str:
+        print("data: {}".format(date_range))
```
enriquezrene, line 114: let's remove this print

fabidick22: Sure, thanks

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -116,21 +125,30 @@ def find_all(
...
         date_range: dict = {},
     ):
-        custom_sql_conditions = []
-        custom_sql_conditions.append(
-            self.create_sql_date_range_filter(date_range)
-        )
+        custom_sql_conditions = [self.create_sql_date_range_filter(date_range)]
 
-        custom_params = self.generate_params(date_range)
+        if event_context.is_admin:
```
Angeluz-07, line 130: Can we move this validation to the DAO?

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -282,19 +300,18 @@ def checks_owner_and_is_not_started(cls, data: dict):
...
-        date_range = self.handle_date_filter_args(args=conditions)
-        time_entries = self.repository.find_all(
-            event_ctx, conditions=conditions, date_range=date_range
-        )
-
-        project_dao = projects_model.create_dao()
-        projects = project_dao.get_all()
-        add_project_name_to_time_entries(time_entries, projects)
-        return time_entries
+        if "start_date" and "end_date" in conditions:
```
Angeluz-07, line 307: I think it would better If we move the logic inside this if to the function `handle_date_filter_args`

commons/data_access_layer/cosmos_db.py
```diff
@@ -222,17 +222,8 @@ def find(
         function_mapper = self.get_mapper_or_dict(mapper)
         return function_mapper(self.check_visibility(found_item, visible_only))
 
-    def find_all(
```
Angeluz-07, line 225: We should latter discuss about formatting. I left a comment only as a reminder of it.

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -74,6 +75,14 @@ def __init__(self, data):  # pragma: no cover
     def running(self):
         return self.end_date is None
 
+    def __add__(self, other):
```
Angeluz-07, line 78: I understand this method can be used to replicate data of projects into a time-entry payload. But I don't see that is being used in the changes of the PR. If you expect to use in a future PR, a #TODO  comment would be very nice to have.

time_tracker_api/time_entries/time_entries_model.py
Angeluz-07: If so, we could check if user is admin and obviate `conditions.update({"owner_id": event_ctx.user_id})` in line 305

time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -116,21 +125,30 @@ def find_all(
...
+        custom_params = self.generate_params(date_range)
+        time_entries = CosmosDBRepository.find_all(
             self,
             event_context=event_context,
             conditions=conditions,
             custom_sql_conditions=custom_sql_conditions,
             custom_params=custom_params,
         )
 
+        if time_entries:
```
Angeluz-07, line 142: Can we move logic from lines 142-148  to an external function ? I think we can apply here https://refactoring.guru/extract-method . That function could go in `time_entries/custom_modules/utils` in a similar way as I did with function `add_project_name_to_time_entries` . The new function could be called `projects_available_for_current_time_entries`

time_tracker_api/time_entries/time_entries_model.py
fabidick22: ok, cool

commons/data_access_layer/cosmos_db.py
fabidick22: fixed up

Issue #160, opened 2020-06-01
Disable automatic clock-outs


enriquezrene:
We implemented a feature to clock out users automatically at midnight. We are facing issues since we are using the time in UTC which is wrong since users are distributed all across the world and for this reason we need to take into account their timezones, for now, let's disable this functionality that will be enabled again after having saved their preferred timezones somewhere.

Angeluz-07:
This is caused by https://github.com/ioet/time-tracker-backend/issues/148

PR #161, opened 2020-06-01
fix: #160 disabling automatic clock-outs

enriquezrene:


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -408,7 +408,7 @@ def find_running(self):
         time_entry = self.repository.find_running(
             event_ctx.tenant_id, event_ctx.user_id
         )
-        self.stop_time_entry_if_was_left_running(time_entry)
+        # self.stop_time_entry_if_was_left_running(time_entry)
```
Angeluz-07, line 411: Can we add little TODO here to add some context to this comment. Something like "Refactor automatic clock-out to consider local timezones". By using the word "refactor" I intend to include the intention to either remove the whole logic around this or extend it taking into account local timezones.

PR #164, opened 2020-06-03
:boom: Feat/retrieve users info from azure#155

### LINKED ISSUE #155: Add required interaction to retrieve user information from Azure
We have an endpoint able to return information about the existing time-entries for different users. The data returned includes user identification (UUID) which does not provide useful information. We need to add interaction with Azure so we can return in this endpoint the user email as well in the payload.

Acceptance Criteria
- The API is able to talk to Azure
- A function to receive the user details using their UUID is added in the API
- The endpoint that returns all the time-entries for reports includes the user email in the payload

Angeluz-07:
Closes #155

### DIFF
--- a/utils/azure_users.py
+++ b/utils/azure_users.py
@@ -1,94 +1,79 @@
 import msal
 import os
 import requests
+from typing import List
 
 
-class MSALConfig:
-    MSAL_CLIENT_ID = os.environ.get('MSAL_CLIENT_ID')
-    MSAL_AUTHORITY = os.environ.get('MSAL_AUTHORITY')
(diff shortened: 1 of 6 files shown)


utils/azure_users.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+        self.config = config
+        self.set_token()
+
+    def set_token(self):
+        response = self.client.acquire_token_for_client(
+            scopes=self.config.SCOPE
+        )
+        if "access_token" in response:
+            self.access_token = response['access_token']
```
fabidick22, line 61: This attribute is not declared in the class (`self.access_token`)

PR #193, opened 2020-07-09
feat: :construction: add time_offset column #190

### LINKED ISSUE #190: Add timezone_offset column on time_entry table
We need to add an additional column as part of the time_entry table. This column is going to save the timezone offset in minutes used during the time_entry creation. 

For example:
-120 if GMT+2
300 if  GMT-5

This column is going to store a numeric value, in the above-mentioned example: -120 or 300 respectively

Angeluz-07:
close #190

### DIFF
--- a/time_tracker_api/time_entries/time_entries_model.py
+++ b/time_tracker_api/time_entries/time_entries_model.py
@@ -83,6 +83,7 @@ class TimeEntryCosmosDBModel(CosmosDBModel):
     technologies: List[str] = field(default_factory=list)
     end_date: str = field(default=None)
     deleted: str = field(default=None)
+    timezone_offset: str = field(default=None)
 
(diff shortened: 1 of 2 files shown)


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -83,6 +83,7 @@ class TimeEntryCosmosDBModel(CosmosDBModel):
     technologies: List[str] = field(default_factory=list)
     end_date: str = field(default=None)
     deleted: str = field(default=None)
+    time_offset: str = field(default=None)
```
enriquezrene, line 86: Let's rename it to timezone_offset please

Angeluz-07: Sure. I'll fix it.

PR #218, opened 2020-10-06
fix: allow limits overlap in time entries #217

### LINKED ISSUE #217: Auto fix times if possible
Clone of https://github.com/ioet/time-tracker-ui/issues/504

Angeluz-07:
Fixes https://github.com/ioet/time-tracker-ui/issues/504
Close #217 

This PR should fix the following problem: 
   You already have a time entry that starts at 11:00:00.
   Then, you want to create a time entry that starts at 9:00:00 and ends at 11:00:00.
   You can't because "There is already a time entry in that range"
   You have to explicitly set your new time entry to (9:00:00 - 10:59:00)

I excluded these cases from the validation that is made when creating/updating time-entries.

Check `time_tracker_api/time_entries/time_entries_model.py` lines 312, 313 for important changes. And `tests/time_tracker_api/time_entries/time_entries_model_test.py ` line 216 for important unit test.

### DIFF
--- a/time_tracker_api/time_entries/time_entries_model.py
+++ b/time_tracker_api/time_entries/time_entries_model.py
@@ -308,6 +308,8 @@ class TimeEntryCosmosDBRepository(CosmosDBRepository):
             SELECT * FROM c
             WHERE ((c.start_date BETWEEN @start_date AND @end_date)
                   OR (c.end_date BETWEEN @start_date AND @end_date))
+                  AND c.start_date!= @end_date
+                  AND c.end_date!= @start_date
             {conditions_clause}
             {ignore_id_condition}
             {visibility_condition}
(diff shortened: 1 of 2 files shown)


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -308,6 +308,8 @@ def find_interception_with_date_range(
             SELECT * FROM c
             WHERE ((c.start_date BETWEEN @start_date AND @end_date)
                   OR (c.end_date BETWEEN @start_date AND @end_date))
+                  AND c.start_date!= @end_date
+                  AND c.end_date!= @start_date
```
rene-enriquez-ct, line 312: ha! this does the trick!!!

rene-enriquez-ct (approved):

PR #235, opened 2020-11-13
229 make admin

magallegos1996:


scastillo-jp (approved):

tests/time_tracker_api/users/users_namespace_test.py
```diff
@@ -23,3 +29,37 @@ def test_users_response_contains_expected_props(
...
     assert 'email' in json.loads(response.data)[0]
     assert 'role' in json.loads(response.data)[0]
+
+
+def test_update_user_role_response_contains_expected_props(
+    client: FlaskClient,
+    valid_header: dict,
+):
+    AzureConnection.update_user_role = Mock(
+        return_value=[{'name': 'dummy', 'email': 'dummy', 'role': 'dummy'}]
```
Angeluz-07, line 39: I think, return  should be a single object here. e.g. : {'name': 'dummy', 'email': 'dummy', 'role': 'dummy'}

tests/time_tracker_api/users/users_namespace_test.py
```diff
@@ -1,8 +1,14 @@
...
 from flask import json
 from flask.testing import FlaskClient
 from flask_restplus._http import HTTPStatus
 from utils.azure_users import AzureConnection
+from faker import Faker
+
+fake = Faker()
+
+valid_user_role_data = {'role': 'admin'}
+user_id = fake.random_int(1, 9999)
```
Angeluz-07, line 11: We could use a fixture instead.

tests/time_tracker_api/users/users_namespace_test.py
```diff
@@ -23,3 +29,37 @@ def test_users_response_contains_expected_props(
...
+    client: FlaskClient,
+    valid_header: dict,
+):
+
+    response = client.put(
+        f'/users/{user_id}', headers=valid_header, json=valid_user_role_data
+    )
+
+    assert HTTPStatus.OK == response.status_code
+    assert valid_user_role_data['role'] == 'admin'
```
Angeluz-07, line 64: I consider this line is not testing.

Angeluz-07 (changes requested):
Good job, just some fixes needed.

Angeluz-07 (approved):
Well done!

scastillo-jp (approved):

PaulRC-ioet (approved):

josepato87 (approved):

time_tracker_api/users/users_namespace.py
```diff
@@ -59,3 +72,16 @@ class User(Resource):
     def get(self, id):
         """Get an user"""
         return {}
+
+    @ns.doc('update_user')
+    @ns.expect(user_input_fields)
+    @ns.response(
+        HTTPStatus.BAD_REQUEST, 'Invalid format or structure of the user'
+    )
+    @ns.marshal_with(user_response_fields)
```
enriquezrene, line 81: I think we need to put this into a separate route. I'd like to have something like
```
/users/{user-id}/roles
```
Then, we can have all the HTTP verbs here like:
```
POST: adds a new roles to the user
DELETE: removes an existing role
...
```

enriquezrene (changes requested):

enriquezrene (approved):

utils/azure_users.py
```diff
@@ -89,4 +74,42 @@ def to_azure_user(item) -> AzureUser:
...
+        there_is_email = len(item['otherMails']) > 0
+        there_is_role = (
+            'extension_1d76efa96f604499acc0c0ee116a1453_role' in item
+        )
+
+        id = item['objectId']
+        name = item['displayName']
+        email = item['otherMails'][0] if there_is_email else ''
+        role = (
+            item['extension_1d76efa96f604499acc0c0ee116a1453_role']
```
juanultimate, line 111: I see this piece of code thrice in this file. It would be great to extract it to a shared place. 
Yo can read the [DRY](https://github.com/HugoMatilla/The-Pragmatic-Programmer#7-the-evils-of-duplication) principle

time_tracker_api/users/users_namespace.py
```diff
@@ -25,17 +25,30 @@
...
         ),
     },
 )
 
 user_response_fields.update(common_fields)
 
+user_input_fields = ns.model(
+    'UserInput',
+    {
+        'role': NullableString(
```
juanultimate, line 42: I am not a restplus expert, but this looks a little bit odd ( this field is **Nullable** and  **required** at the same time)
```
 'role': NullableString(
            title="User's Role",
            required=True,
            max_length=50,
            description='Role assigned to the user by the tenant',
            example=faker.word(['time-tracker-admin']),
        ),
```
Would not it better if we make this field optional??

juanultimate (changes requested):

utils/azure_users.py
juanultimate: In addition I would extract  the 1d76efa96f604499acc0c0ee116a1453 hash as a env variable since it eventually might change.

time_tracker_api/users/users_namespace.py
Angeluz-07: we'll use this field to allow PUT calls to switch user's roles:
with input :
{'role': 'time-tracker-admin'} we give the role

with input : 
{'role': null} we remove the role

for that reason is nullable & string, and it is required because this is the field we want to change for this endpoint. 

Does that make sense?

utils/azure_users.py
Angeluz-07: I moved into a variable of the class. May I argue against use it as a env variable? I consider that as a variable is enough.

time_tracker_api/users/users_namespace.py
juanultimate: Yes, it does. 
However, since the URI for this endpoint for this resource is `/users/<user-id>/roles` you are going to add roles ( POST operation) and remove roles ( DELETE operation), I would suggest using the POST and DELETE operations instead of PUT. 
It would be different if the URI were `/users/<user-id>`, in that case I would use the PUT operation.

time_tracker_api/users/users_namespace.py
```diff
@@ -25,17 +25,30 @@
...
             description='Role assigned to the user by the tenant',
-            example=faker.word(['admin']),
+            example=faker.word(['time-tracker-admin']),
         ),
     },
 )
 
 user_response_fields.update(common_fields)
 
+user_input_fields = ns.model(
```
juanultimate, line 39: Should not this be named something like `role_input_fields` instead of `user_input_fields `??
I see you are only sending the role attribute

Angeluz-07: Great, I think user_role_input_fields would be better, because is only about user_role.

time_tracker_api/users/users_namespace.py
Angeluz-07: @juanultimate  We used PUT because looking at our code base, this requirement it looked pretty similar to an update operation(In our codebase all updates are managed via PUT).

I understand your point, probably from the perspective of a UI dev, it is weird to use PUT in something that has no entity. Maybe a POST would seem more natural. About the DELETE operation, not that sure. We may need to discuss this in a meeting.

time_tracker_api/users/users_namespace.py
```diff
@@ -17,45 +20,72 @@
...
+        HTTPStatus.BAD_REQUEST, 'Invalid format or structure of the user'
+    )
+    @ns.marshal_with(user_response_fields)
+    def post(self, id):
+        """Create user's role"""
+        return azure_connection.update_user_role(id, ns.payload['role'])
+
+
+@ns.route('/<string:user_id>/roles/<string:role_id>')
+@ns.response(HTTPStatus.NOT_FOUND, 'User not found')
```
juanultimate, line 82: User not found? IMO, it is better to say `Role not found`

time_tracker_api/users/users_namespace.py
```diff
@@ -17,45 +20,72 @@
...
         """List all users"""
-        from utils.azure_users import AzureConnection
-
-        azure_connection = AzureConnection()
         return azure_connection.users()
 
 
-@ns.route('/<string:id>')
+@ns.route('/<string:id>/roles')
 @ns.response(HTTPStatus.NOT_FOUND, 'User not found')
```
juanultimate, line 66: User not found? I would say it is better to say `Role not found`

time_tracker_api/users/users_namespace.py
```diff
@@ -17,45 +20,72 @@
...
 @ns.response(HTTPStatus.UNPROCESSABLE_ENTITY, 'The id has an invalid format')
 @ns.param('id', 'The user identifier')
-class User(Resource):
-    @ns.doc('get_user')
+class UserRoles(Resource):
+    @ns.doc('create_user_role')
+    @ns.expect(user_role_input_fields)
+    @ns.response(
+        HTTPStatus.BAD_REQUEST, 'Invalid format or structure of the user'
```
juanultimate, line 73: Invalid format or structure of the user? Is not this for roles?

juanultimate (approved):

Issue #227, opened 2020-11-07
Remove static methods from TimeEntriesDao


enriquezrene:
The class TimeEntriesDao has the following static method:
```
@staticmethod
def current_user_id():
        return current_user_id()
```

As you can see, it returns an existing function part of the security.py file. We can use directly that function instead of returning the same function. Please refactor this, we should aim for having 0 static methods. We should use either a POO approach or functions directly. But it does not make any sense to have static methods.

PR #237, opened 2020-11-13
fix: remove static method from TimeEntriesDao #227

gqroot:


time_tracker_api/time_entries/time_entries_model.py
```diff
@@ -499,7 +496,11 @@ def update(self, id, data: dict, description=None):
         time_entry = self.repository.find(id, event_ctx)
         self.check_whether_current_user_owns_item(time_entry)
```
enriquezrene, line 497: Here we have a usage of this method using `self.` as prefix which AFAIK it means it is going to use the method which is part of the class instance, right?

If so, why should this method be static, I think we can get rid of the `@staticmethod` annotation or  am I wrong?

```
    @staticmethod
    def check_whether_current_user_owns_item(data):
        if data.owner_id is not None and data.owner_id != current_user_id():
            raise CustomError(
                HTTPStatus.FORBIDDEN,
                "The current user is not the owner of this time entry",
            )
```

enriquezrene (changes requested):
Please address my comment

time_tracker_api/time_entries/time_entries_model.py
gqroot: I agree, in the first push the test fell, so I left it like that, now I tried again and the test no longer fell, so I will leave it as suggested.

enriquezrene (approved):

scastillo-jp (approved):

PR #247, opened 2020-12-16
Revert "Revert "TT-69 fix: adding or editing an entry must not overla…

gqroot:
…p the date of any other entry""

This reverts commit 26ddb20ea03dc69b5e60d70a8c1a8bb0d5f49126.


josepato87 (approved):
I think it's ok, if you could think in other tests cases that could be used in the current data providers, would be nice. ;)

PR #249, opened 2021-01-05
TT-94 update role support

Angeluz-07:


PaulRC-ioet (approved):
For my part, I do not find any errors or anything that you should change, but in the tests, we should do tests that fail just to test those scenarios?

israteneda (approved):

scastillo-jp (approved):

PR #250, opened 2021-01-08
Tt 98 toogles backend

PaulRC-ioet:
Implementation of feature toggles in the backend.


time_tracker_api/time_entries/time_entries_namespace.py
```diff
@@ -267,6 +268,24 @@ def get(self):
         return time_entries_dao.get_lastest_entries_by_project(conditions={})
 
 
+@ns.route('/feature-toggles')
+class FeaturesToggles(Resource):
```
Angeluz-07, line 272: Not sure if this piece of code (lines 271 to 288)should be included.  You already tested, your code works. I consider that we could remove this and put this whole example in the JIRA issue(as an example of how to use the current implementation).

e.g.:
```
ftm = FeatureToggleManager('<my-feature-flag>')
is_enabled = ftm.is_toggle_enabled_for_user()
if is_enabled:
    # My logic if toggle is enabled....
else:
   # My logic if toggle is disabled....
```

tests/commons/feature_toggles/feature_toggles_manager_test.py
```diff
@@ -0,0 +1,63 @@
+from commons.feature_toggles.feature_toggle_manager import FeatureToggleManager
+from unittest.mock import Mock, patch
+from pytest import mark
+
+
+def mock_payload(enabled, user):
```
Angeluz-07, line 6: Maybe, we could change the name of the function to `mock_feature_toggle_config_response` or something similar. `mock_payload` it seems too general to me.

commons/feature_toggles/feature_toggle_manager.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_configuration(self, key: str, label: str):
+        configuration = self.client.get_configuration_setting(
+            key=f".appconfig.featureflag/{self.key}", label=self.label
+        )
+
+        return configuration
+
+    def get_data_configuration(self):
+        self.configuration = self.get_configuration(self.key, self.label)
```
Angeluz-07, line 46: I think there's no need to save `configuration` as property. As we always want the result on-demand. We could only save it in a variable inside this function.

commons/feature_toggles/feature_toggle_manager.py
```diff
@@ -0,0 +1,69 @@
...
+        connection_str = self.config.AZURE_APP_CONFIGURATION_CONNECTION_STRING
+        client = AzureAppConfigurationClient.from_connection_string(
+            connection_str
+        )
+
+        return client
+
+    def get_configuration(self, key: str, label: str):
+        configuration = self.client.get_configuration_setting(
+            key=f".appconfig.featureflag/{self.key}", label=self.label
```
Angeluz-07, line 40: We are using `self.key` and `self.label` variables in the function. But we are also passing key, and label as arguments in the function call.

We could use `self.var` or the `var` . I would suggest use argument vars.

commons/feature_toggles/feature_toggle_manager.py
```diff
@@ -0,0 +1,69 @@
+import os
+import json
+from time_tracker_api.security import current_user_email
+from azure.appconfiguration import AzureAppConfigurationClient
+
+
+class MSConfig:
```
Angeluz-07, line 7: `MSConfig` It was because this class was used with MS variables. In this case is different, we could use something like `FeatureToggleConfig` or `FTConfig` for example.

commons/feature_toggles/feature_toggle_manager.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+
+class FeatureToggleManager:
+    def __init__(self, key: str, label: str = None, config=MSConfig):
+        self.key = key
+        self.label = label
+        self.config = config
+        self.client = self.get_azure_app_configuration_client()
+        self.configuration = {}
```
Angeluz-07, line 28: I consider we could get rid of this field. As I mentioned in another comment https://github.com/ioet/time-tracker-backend/pull/250/files#r554218908

commons/feature_toggles/feature_toggle_manager.py
```diff
@@ -0,0 +1,69 @@
...
+    def check_variables_are_defined():
+        auth_variables = ['AZURE_APP_CONFIGURATION_CONNECTION_STRING']
+        for var in auth_variables:
+            if var not in os.environ:
+                raise EnvironmentError(
+                    "{} is not defined in the environment".format(var)
+                )
+
+    check_variables_are_defined()
+    AZURE_APP_CONFIGURATION_CONNECTION_STRING = os.environ.get(
```
Angeluz-07, line 17: I know this is not in your ticket scope. But if you find some way to improve this class, let us know.

Angeluz-07 (commented):
Some comments. I must say that we need to also update the `.env.example` file with our new env variable.

commons/feature_toggles/feature_toggle_manager.py
PaulRC-ioet: We have changed our class that function, Roberto is going to help us to refactor this specific class, hopefully, it will be ready today if not we will create a ticket.

commons/feature_toggles/feature_toggle_manager.py
PaulRC-ioet: It has been deleted that field.

commons/feature_toggles/feature_toggle_manager.py
PaulRC-ioet: We have changed to `FeatureToggleConfig` because is more explicit

commons/feature_toggles/feature_toggle_manager.py
PaulRC-ioet: We already change to use the argument `vars` in the function.

commons/feature_toggles/feature_toggle_manager.py
PaulRC-ioet: it's changed, thanks.

tests/commons/feature_toggles/feature_toggles_manager_test.py
PaulRC-ioet: We have changed the name to `mock_feature_toggle_config_response`

time_tracker_api/time_entries/time_entries_namespace.py
PaulRC-ioet: Ok, we have moved this example code to JIRA.

time_tracker_api/time_entries/time_entries_namespace.py
```diff
@@ -17,6 +17,7 @@
     remove_required_constraint,
 )
 from time_tracker_api.time_entries.time_entries_dao import create_dao
+from commons.feature_toggles.feature_toggle_manager import FeatureToggleManager
```
Angeluz-07, line 20: We need to remove this line also.

PaulRC-ioet: Done.

Angeluz-07 (approved):

PR #251, opened 2021-01-13
TT - 112 Update role field

Angeluz-07:
See TT - 112 for details.

### DIFF
--- a/tests/time_tracker_api/users/users_namespace_test.py
+++ b/tests/time_tracker_api/users/users_namespace_test.py
@@ -6,22 +6,6 @@ from utils.azure_users import AzureConnection
 from pytest import mark
 
 
-def test_users_response_contains_expected_props(
-    client: FlaskClient, valid_header: dict,
-):
-
-    AzureConnection.users = Mock(
-        return_value=[{'name': 'dummy', 'email': 'dummy', 'role': 'dummy'}]
-    )
-
-    response = client.get('/users', headers=valid_header,)
-
-    assert HTTPStatus.OK == response.status_code
-    assert 'name' in json.loads(response.data)[0]
-    assert 'email' in json.loads(response.data)[0]
-    assert 'role' in json.loads(response.data)[0]
-
-
 @patch(
     'commons.feature_toggles.feature_toggle_manager.FeatureToggleManager.get_azure_app_configuration_client'
 )
@@ -48,6 +32,8 @@ def test_feature_toggle_is_on_then_role_field_is_list(
     users_v2_mock.assert_called()
     users_mock.assert_not_called()
     assert HTTPStatus.OK == response.status_code
+    assert 'name' in json.loads(response.data)[0]
+    assert 'email' in json.loads(response.data)[0]
     assert 'roles' in json.loads(response.data)[0]
     assert ['dummy-role'] == json.loads(response.data)[0]['roles']
 
@@ -78,6 +64,8 @@ def test_feature_toggle_is_off_then_role_field_is_string(
     users_mock.assert_called()
     users_v2_mock.assert_not_called()
     assert HTTPStatus.OK == response.status_code
+    assert 'name' in json.loads(response.data)[0]
+    assert 'email' in json.loads(response.data)[0]
     assert 'role' in json.loads(response.data)[0]
     assert 'dummy-role' == json.loads(response.data)[0]['role']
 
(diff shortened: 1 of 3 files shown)


utils/azure_users.py
```diff
@@ -119,6 +146,21 @@ def to_azure_user(self, item) -> AzureUser:
...
+
+        id = item['objectId']
+        name = item['displayName']
+        email = item['otherMails'][0] if there_is_email else ''
+        roles = []
+        for role_id in ROLE_FIELD_VALUES.keys():
+            field_name, _ = ROLE_FIELD_VALUES[role_id]
+            if field_name in item:
+                field_value = item[field_name]
+                roles.append(field_value)
```
josepato87, line 160: Can we use something more sophisticated like `filter` or `reduce`, maybe `lamba functions` or `comprehension lists`

time_tracker_api/users/users_namespace.py
```diff
@@ -57,7 +66,11 @@ class Users(Resource):
     @ns.marshal_list_with(user_response_fields)
     def get(self):
         """List all users"""
-        return AzureConnection().users()
+        ftm = FeatureToggleManager('bk-user-role-field')
```
josepato87, line 69: I would suggest to change `ftm` to the name of the FF name plus the word Toggle.

josepato87 (commented):
Just left some easy comments

tests/time_tracker_api/users/users_namespace_test.py
```diff
@@ -6,20 +6,68 @@
...
 
+@patch(
+    'commons.feature_toggles.feature_toggle_manager.FeatureToggleManager.get_azure_app_configuration_client'
+)
+@patch(
+    'commons.feature_toggles.feature_toggle_manager.FeatureToggleManager.is_toggle_enabled_for_user'
+)
+@patch('utils.azure_users.AzureConnection.users')
+@patch('utils.azure_users.AzureConnection.users_v2')
+def test_feature_toggle_is_off_then_role_field_is_string(
```
israteneda, line 49: Is it possible to simplify the test to execute the two functions in one, since the code is almost the same in the two functions? Is it possible to parameterize the mocking of the users? It's just a suggestion.

time_tracker_api/users/users_namespace.py
```diff
@@ -57,7 +66,11 @@ class Users(Resource):
...
+        if ftm.is_toggle_enabled_for_user():
+            return AzureConnection().users_v2()
+        else:
+            return AzureConnection().users()
```
josepato87, line 73: Maybe we could avoid that `else` by the short circuit that the return inside the `if` clause is doing. :)

tests/time_tracker_api/users/users_namespace_test.py
Angeluz-07: Yes the problem to parametrize is this code:
```
Test 1
users_mock.assert_called()  
users_v2_mock.assert_not_called()

Test 2
users_mock.assert_not_called()
users_v2.assert_called()
``` 
But, I did not find a convenient way to parametrize that.

utils/azure_users.py
```diff
@@ -119,6 +146,21 @@ def to_azure_user(self, item) -> AzureUser:
         role = item[self.role_field] if there_is_role else None
         return AzureUser(id, name, email, role)
 
+    def to_azure_user_v2(self, item) -> AzureUser:
```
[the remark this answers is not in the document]
Angeluz-07: Yes. I 'll do it.

utils/azure_users.py
Angeluz-07: Sure :)

time_tracker_api/users/users_namespace.py
Angeluz-07: It is ok, can be reduced.

time_tracker_api/users/users_namespace.py
Angeluz-07: Agree.

Angeluz-07: It seemed a good idea first, but actually the name of the variable became long.
```
bk_user_role_field_toggle = FeatureToggleManager('bk-user-role-field')
if bk_user_role_field_toggle.is_toggle_enabled_for_user():
    return AzureConnection().users_v2()
return AzureConnection().users()
```
Does that seem right to you? I think we can come up with a shorter name.

utils/azure_users.py
Angeluz-07: Thanks for the suggestion. I found two ways to do it 
```
field_names = [field_name for (field_name, _)  in ROLE_FIELD_VALUES.values()]
field_names_in_item = filter(lambda field_name: field_name in item , field_names)
roles = map(lambda field_name: item[field_name], field_names_in_item)
roles = list(roles)

or

roles = [
    item[field_name] for (field_name, field_value) in ROLE_FIELD_VALUES.values()
    if field_name in item
]

```

I think the second one in cleaner.

israteneda (approved):

time_tracker_api/users/users_namespace.py
Angeluz-07: changed to `user_role_field_toggle`

josepato87 (approved):
LGTM

PR #252, opened 2021-01-22
Tt 123 separe test users in report data

PaulRC-ioet:
Feature to filter time entries report by test and admin roles

### DIFF
--- a/tests/utils/azure_users_test.py
+++ b/tests/utils/azure_users_test.py
@@ -29,8 +29,7 @@ def test_azure_connection_is_test_user(
 @patch('msal.ConfidentialClientApplication', Mock())
 @patch('utils.azure_users.AzureConnection.get_token', Mock())
 @patch('requests.get')
-def test_azure_connection_get_test_user_ids(get_mock,):
-
+def test_azure_connection_get_test_user_ids(get_mock):
     response_mock = Mock()
     response_mock.status_code = 200
     response_mock.json = Mock(
--- a/tests/utils/azure_users_test.py
+++ b/tests/utils/azure_users_test.py
@@ -1,0 +1,44 @@
+from unittest.mock import Mock, patch
+from utils.azure_users import AzureConnection, ROLE_FIELD_VALUES, AzureUser_v2
+from pytest import mark
+
+
+@patch('msal.ConfidentialClientApplication')
+@patch('utils.azure_users.AzureConnection.get_token')
+@patch('utils.azure_users.AzureConnection._get_user')
+@mark.parametrize(
+    'field_name,field_value,expected',
+    [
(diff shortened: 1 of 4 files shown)


Angeluz-07 (approved):

utils/azure_users.py
```diff
@@ -185,3 +186,32 @@ def get_role_data(self, role_id, is_grant=True):
...
+            endpoint=self.config.ENDPOINT,
+            field_name=field_name,
+            field_value=field_value,
+        )
+        response = requests.get(endpoint, auth=BearerAuth(self.access_token))
+        assert 200 == response.status_code
+        assert 'value' in response.json()
+        return response.json()['value']
+
+    def get_test_user_ids(self):
```
josepato87, line 215: I just left the same comment in Robert's PR. Why do we have this "name duplicated" method

time_tracker_api/time_entries/time_entries_dao.py
```diff
@@ -93,21 +94,35 @@ def build_custom_query(self, is_admin: bool, conditions: dict = None):
     def get_all(self, conditions: dict = None, **kwargs) -> list:
         event_ctx = self.create_event_context("read-many")
         conditions.update({"owner_id": event_ctx.user_id})
-
+        is_complete_query = conditions.get("user_id") == '*'
```
josepato87, line 97: I'll move this variable next to its usage, before line 116

tests/utils/azure_users_test.py
```diff
@@ -0,0 +1,44 @@
+from unittest.mock import Mock, patch
+from utils.azure_users import AzureConnection, ROLE_FIELD_VALUES, AzureUser_v2
+from pytest import mark
+
+
+@patch('msal.ConfidentialClientApplication')
+@patch('utils.azure_users.AzureConnection.get_token')
+@patch('utils.azure_users.AzureConnection._get_user')
+@mark.parametrize(
+    'field_name,field_value,expected',
```
josepato87, line 10: Maybe we could change "expected" to "expectedTestUser"

time_tracker_api/time_entries/time_entries_dao.py
PaulRC-ioet: Sure, that was my first option, but this variable must be declared here because after it goes through `custom_query` method and these values ​​are lost.

utils/azure_users.py
PaulRC-ioet: With Roberto, we have rethought it and we have united it in a single method so as not to have these repeated functions.

tests/utils/azure_users_test.py
PaulRC-ioet: Sure, we have changed to `is_test_user_expected_value` it's a more descriptive name.

josepato87 (approved):

PR #255, opened 2021-01-26
fix: :wastebasket: TT-131 remove feature toggle for user role field

Angeluz-07:


israteneda (approved):

PaulRC-ioet (approved):

tests/time_tracker_api/users/users_namespace_test.py
```diff
@@ -2,99 +2,33 @@
 from flask import json
 from flask.testing import FlaskClient
 from flask_restplus._http import HTTPStatus
-from utils.azure_users import AzureConnection
```
scastillo-jp, line 5: I have a question if this line `from utils.azure_users import AzureConnection` is deleted, does this line 8 work?  `@patch('utils.azure_users.AzureConnection.get_msal_client', Mock())`

Angeluz-07: Yes, It works. The patch method makes the call to the function.

PR #258, opened 2021-02-12
refactor: TT-114 Refactor in the validation of environment variables class…

LEON12699:
Refractor validation of environment variables


utils/azure_users.py
```diff
@@ -3,24 +3,20 @@
...
-            'MS_SECRET',
-            'MS_SCOPE',
-            'MS_ENDPOINT',
-        ]
-        for var in auth_variables:
-            if var not in os.environ:
-                raise EnvironmentError(
-                    "{} is not defined in the environment".format(var)
-                )
+    auth_variables = [
```
Angeluz-07, line 10: lets change `auth_variables` for `ms_variables`

commons/feature_toggles/feature_toggle_manager.py
```diff
@@ -2,19 +2,10 @@
 import json
 from time_tracker_api.security import current_user_email
 from azure.appconfiguration import AzureAppConfigurationClient
-
+from utils.check_variables import  check_variables_are_defined
```
israteneda, line 5: Only a little change, it seems we have an extra space in this part `import  check_variables_are_defined`

requirements/time_tracker_api/dev.txt
```diff
@@ -15,4 +15,7 @@ pytest-mock==2.0.0
 coverage==4.5.1
 
 # Git hooks
-pre-commit==2.2.0
\ No newline at end of file
+pre-commit==2.2.0
+
+#Azure functions
+azure-functions==1.6.0
```
Angeluz-07, line 21: This library is included when you run `pip install -r requirements/time-tracker-events/dev.txt`. No need to put it here. Maybe we should update the Readme to mention this, but that would be another ticket.

commons/feature_toggles/feature_toggle_manager.py
israteneda: Also, I think we could change the file `check_variables` to something more generic to put other useful functions in there, something like `enviroment_variables`. If we don't have a more generic name for the file, the current name could be stayed.

utils/check_variables.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+def check_variables_are_defined(variables):
+  for var in variables:
+    if var not in os.environ:
```
israteneda, line 5:
```suggestion
  value = os.environ.get(variable)
    if value is None:
```
We can analyze if this is a better solution, since `if var not in os.environ:` is looping for all environment variables, maybe `value = os.environ.get(variable)` is more optimal. We could discuss this.

requirements/time_tracker_api/dev.txt
LEON12699: I didn't see what time tracker events is included in it, I'm going to remove it. But I think it is important to mention it in another ticket to update the README

commons/feature_toggles/feature_toggle_manager.py
LEON12699: It is a good observation I will correct it

requirements/time_tracker_api/dev.txt
Angeluz-07: The call to `pip install -r requirements/time-tracker-events/dev.txt` makes a call to `requirements/time-tracker-events/prod.txt` which makes another call to `-r ../../time_tracker_events/requirements.txt` that is where this library is included.

Angeluz-07 (approved):

requirements/time_tracker_api/dev.txt
LEON12699: You are right, thanks for the observation

israteneda (approved):

PR #259, opened 2021-02-18
feat: TT-154 add is user member of a group endpoint

PaulRC-ioet:
Endpoint to check if a user is member of a group in azure.

### DIFF
--- a/utils/azure_users.py
+++ b/utils/azure_users.py
@@ -169,3 +169,37 @@ class AzureConnection:
         assert 200 == response.status_code
         assert 'value' in response.json()
         return [item['objectId'] for item in response.json()['value']]
+
+    def get_group_id_by_group_name(self, group_name):
+        endpoint_get_groups = "{endpoint}/groups?api-version=1.6&$select=objectId&$filter=displayName eq '{group_name}'".format(
+            endpoint=self.config.ENDPOINT, group_name=group_name
+        )
+
+        response_get_groups = requests.get(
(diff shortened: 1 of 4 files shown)


utils/azure_users.py
```diff
@@ -169,3 +169,37 @@ def get_test_user_ids(self):
         assert 200 == response.status_code
         assert 'value' in response.json()
         return [item['objectId'] for item in response.json()['value']]
+
+    def get_group_id_by_group_name(self, group_name):
+        endpoint_get_groups = "{endpoint}/groups?api-version=1.6&$select=objectId&$filter=displayName eq '{group_name}'".format(
```
Angeluz-07, line 174: I think we could change the name of variable from `endpoint_get_groups` to `endpoint`. Also `response_get_groups` to `response`

utils/azure_users.py
```diff
@@ -169,3 +169,37 @@ def get_test_user_ids(self):
         assert 200 == response.status_code
         assert 'value' in response.json()
         return [item['objectId'] for item in response.json()['value']]
+
+    def get_group_id_by_group_name(self, group_name):
```
Angeluz-07, line 173: Please add a type hinting here. `-> str`

utils/azure_users.py
PaulRC-ioet: Sorry i forgot to change that variables, now i have changed it.

Angeluz-07 (approved):

PR #263, opened 2021-03-03
feat: TT-169 add enpoints to add/remove users to/from groups

Angeluz-07:


tests/utils/azure_users_test.py
```diff
@@ -176,3 +176,45 @@ def test_get_groups_and_users_called_once_by_instance(
...
+    get_group_id_by_group_name_mock.return_value = 'dummy_group'
+
+    response_mock = Mock()
+    response_mock.status_code = 204
+    post_mock.return_value = response_mock
+
+    azure_connection = AzureConnection()
+    azure_connection.add_user_to_group('dummy_user_id', 'dummy_group')
+
+    get_group_id_by_group_name_mock.assert_called_once()
```
PaulRC-ioet, line 198: Do you think you can send an assert with the mock of a user in the test that is changing the group? It is optional because I see that if you are testing the methods.

PaulRC-ioet (commented):
Good job, with this it would be one more step to migrate to the groups. Everything's fine.

israteneda (approved):

scastillo-jp (approved):

PaulRC-ioet (approved):

PR #265, opened 2021-03-16
Tt 180 create method get activities by id list in activitydao

kellycastrof:
Create methods find_all_with_id_in_list in ActivityCosmosDBRepository and get_all_with_id_in_list
Implementation of the test methods in activities_model_test.py in tests\time_tracker_api\activities

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -87,7 +87,11 @@ def cosmos_db_model():
     return {
         'id': 'test',
         'partition_key': PartitionKey(path='/tenant_id'),
-        'unique_key_policy': {'uniqueKeys': [{'paths': ['/email']},]},
+        'unique_key_policy': {
+            'uniqueKeys': [
+                {'paths': ['/email']},
+            ]
+        },
     }
 
 
@@ -230,13 +234,6 @@ def activity_repository(app: Flask) -> ActivityCosmosDBRepository:
             init_app(app)
 
     return ActivityCosmosDBRepository()
-
-
-@pytest.fixture
-def activities_dao():
-    from time_tracker_api.activities.activities_model import create_dao
-
-    return create_dao()
 
 
 @pytest.fixture(scope="session")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,6 @@
 from datetime import datetime, timedelta
 
 import jwt
-import base64
 import pytest
 from faker import Faker
 from flask import Flask
@@ -88,11 +87,7 @@ def cosmos_db_model():
     return {
         'id': 'test',
         'partition_key': PartitionKey(path='/tenant_id'),
-        'unique_key_policy': {
-            'uniqueKeys': [
-                {'paths': ['/email']},
-            ]
-        },
+        'unique_key_policy': {'uniqueKeys': [{'paths': ['/email']},]},
     }
 
 
@@ -239,9 +234,7 @@ def activity_repository(app: Flask) -> ActivityCosmosDBRepository:
 
 @pytest.fixture
 def activities_dao():
-    from time_tracker_api.activities.activities_model import (
-        create_dao,
-    )
+    from time_tracker_api.activities.activities_model import create_dao
 
     return create_dao()
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 
 import jwt
+import base64
 import pytest
 from faker import Faker
 from flask import Flask
@@ -13,6 +14,9 @@ from time_tracker_api.database import init_sql
 from time_tracker_api.security import get_or_generate_dev_secret_key
 from time_tracker_api.time_entries.time_entries_repository import (
     TimeEntryCosmosDBRepository,
+)
+from time_tracker_api.activities.activities_model import (
+    ActivityCosmosDBRepository,
 )
 
 fake = Faker()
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -222,6 +221,24 @@ def running_time_entry(
...
+        from commons.data_access_layer.cosmos_db import init_app, cosmos_helper
+
+        if cosmos_helper is None:
+            init_app(app)
+
+    return ActivityCosmosDBRepository()
+
+
+@pytest.fixture
+def activities_dao():
```
Angeluz-07, line 236: We are not using the activities dao,we may remove it.

time_tracker_api/activities/activities_model.py
```diff
@@ -39,12 +41,72 @@ def __str___(self):
...
+    def convert_list_to_tuple_string(self, id_list):
+        self.validate_list(id_list)
+        id_value = (
+            f"('{id_list[0]}')" if len(id_list) == 1 else str(tuple(id_list))
+        )
+        return id_value
+
+    def validate_list(self, id_list):
+        assert isinstance(id_list, list)
+        assert len(id_list) > 0
```
israteneda, line 67: These methods seem generic, what if you move them to the `utils` folder to be able to use them in other classes, do you think it is possible or do they use a specific logic for this class?

kellycastrof: It's a nice idea, @eduardisrael will implement something similar, so he will need that functions.

kellycastrof: https://github.com/ioet/time-tracker-backend/blob/c94b511ba5720753b273fd87fb6bb2bafb4e8851/utils/repository.py#L1-L11

@israteneda  now it is implemented!

thegreatyamori (approved):

tests/conftest.py
kellycastrof: I just deleted it

VanessaIniguez (approved):

israteneda (approved):

utils/repository.py
```diff
@@ -0,0 +1,11 @@
+def convert_list_to_tuple_string(id_list):
+    validate_list(id_list)
```
scastillo-jp, line 2: validate_list(id_list): Does this get a list of ids? but the output of that function is completely ignored.
validate_list: should validate the list but it returns something

id_list[0]: I think you can use pop to get the firs value or use a map to reduce the list.
id_list rename for ids_list

I think this can be improved:
f"('{ids_list[0]}')" if len(ids_list) == 1 else str(tuple(ids_list))
for
[str(id) for id in ids_list]

or if you want tuples
(str(id) for id in ids_list)

kellycastrof: validate_list doesn't return a list, it only does assertions to validate if the id list is valid or not  -> if it is a list and if it has elements, because of that the output is ignored.
What I need is not the tuple, but the string of a tuple. For example '("1", "2")' instead of ("1", "2") 

When I only have one element and convert it to a tuple it would have something similar to ("1",) and this gives an error in the database because of the comma, for this reason,  maybe to make it better understood I could do something similar to this 
https://github.com/ioet/time-tracker-backend/blob/b095f56eea0509f56211724962449fbac3d53b8b/utils/extend_model.py#L69-L71

kellycastrof: I renamed id_list to ids_list in utils/repository.py 
https://github.com/ioet/time-tracker-backend/blob/79901cec1c2a39d2f5c4d49e6b6f426cc2bfb454/utils/repository.py#L1-L2

Also, I renamed in activity_model.py id_list to activities_ids_list 
https://github.com/ioet/time-tracker-backend/blob/79901cec1c2a39d2f5c4d49e6b6f426cc2bfb454/time_tracker_api/activities/activities_model.py#L66
And change the f"('{ids_list[0]}')" if len(ids_list) == 1 else str(tuple(ids_list)) to this
https://github.com/ioet/time-tracker-backend/blob/79901cec1c2a39d2f5c4d49e6b6f426cc2bfb454/utils/repository.py#L3-L6
I think with that changes it is more readability, the point of that is get a string of the tuple, without extra comma

kellycastrof: @scastillo-jp

time_tracker_api/activities/activities_model.py
```diff
@@ -39,12 +46,64 @@ def __str___(self):
...
-            CosmosDBDao.__init__(self, repository)
 
-    return ActivityCosmosDBDao()
+class ActivityCosmosDBDao(APICosmosDBDao, ActivityDao):
+    def __init__(self, repository):
+        CosmosDBDao.__init__(self, repository)
+
+    def get_all_with_id_in_list(
+        self,
+        activities_ids_list,
```
Angeluz-07, line 96: `activity_ids_list` or `activity_id_list` or `activity_ids`  ? I think the current one is a bit long.

time_tracker_api/activities/activities_model.py
```diff
@@ -39,12 +46,64 @@ def __str___(self):
...
+    ):
+        event_ctx = self.create_event_context("read-many")
+        activities_list = self.repository.find_all_with_id_in_list(
```
Angeluz-07, line 99: We could return the value directly.

utils/repository.py
```diff
@@ -0,0 +1,13 @@
+def convert_list_to_tuple_string(ids_list):
+    validate_list(ids_list)
+    id_value = (
```
Angeluz-07, line 3: You could use `result` as a temp variable to be returned.

utils/repository.py
```diff
@@ -0,0 +1,13 @@
...
+        str(tuple(ids_list)).replace(",", "")
+        if len(ids_list) == 1
+        else str(tuple(ids_list))
+    )
+    return id_value
+
+
+def validate_list(ids_list):
```
Angeluz-07, line 11: Given that we have this two function without context. We could put the asserts in the function `convert_list_to_tuple_string` and make only one function. As they are used only there. What do you think?

time_tracker_api/activities/activities_model.py
kellycastrof: Maybe activities_ids is better, activities_ids: List str  for more details

time_tracker_api/activities/activities_model.py
kellycastrof: Yeah, but I think that it is more legible in that way.

utils/repository.py
kellycastrof: result instead of id_value?

utils/repository.py
kellycastrof: I'm not sure, I think in that way the function would be doing more things than it should.

VanessaIniguez (approved):

Angeluz-07 (approved):
Good job.

PR #266, opened 2021-03-17
TT-181 create method to find all projects with list of customer ids

eduardisrael:
create find_all_with_customer_id_in_list method and unit tests

### DIFF
--- a/time_tracker_api/projects/projects_model.py
+++ b/time_tracker_api/projects/projects_model.py
@@ -66,7 +66,7 @@ class ProjectCosmosDBRepository(CosmosDBRepository):
             mapper=ProjectCosmosDBModel,
         )
 
-    def create_sql_in_condition(self, customer_ids_list):
+    def create_sql_customer_id_in_condition(self, customer_ids_list):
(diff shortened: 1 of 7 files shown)


time_tracker_api/projects/projects_model.py
```diff
@@ -63,6 +66,36 @@ def __init__(self):
...
+        visible_only=True,
+        mapper: Callable = None,
+    ):  
+        visibility = self.create_sql_condition_for_visibility(visible_only)
+        query_str = """
+            SELECT * FROM c
+            WHERE {condition}
+            {visibility_condition}
+            """.format(
+            condition=self.create_sql_in_condition(customer_ids_list),
```
Angeluz-07, line 87: We could call this function `create_sql_customer_id_in_condition` and use it with a variable for better readability.

Angeluz-07: Or better talk with @kellycastrof, she thinks this may be a more generic function. You could try too.

Angeluz-07:
Deprecated as it was created another PR in https://github.com/ioet/time-tracker-backend/pull/267/files

PR #269, opened 2021-03-24
refactor: TT-187 replace sql_in_condition & create_sql_customer_id_in

eduardisrael:
Replace sql_in_condition & create_sql_customer_id_in_condition functions for reusable function.

### DIFF
--- a/utils/repository.py
+++ b/utils/repository.py
@@ -8,7 +8,7 @@ def convert_list_to_tuple_string(ids_list):
     )
     return result
 
-def create_sql_in_condition(values, field):
+def create_sql_in_condition(field, values):
     tuple_string = convert_list_to_tuple_string(values)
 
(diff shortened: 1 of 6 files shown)


utils/repository.py
```diff
@@ -7,3 +7,8 @@ def convert_list_to_tuple_string(ids_list):
         else str(tuple(ids_list))
     )
     return result
+
+def create_sql_in_condition(values, field):
```
kellycastrof, line 11: Maybe, it is better to call first the name of the field and then the list
`def create_sql_in_condition(field, values)`
If you have many elements in the list, the name of the field could be lost

kellycastrof (approved):

thegreatyamori (approved):

Angeluz-07 (approved):

PR #270, opened 2021-03-26
Tt 185 find all method in time entries

kellycastrof:
**CosmosDBQueryBuilder** and **TimeEntryQueryBuilder** class implementation, 
new function in utils/repository.py called get_string_without_empty_spaces which is used in tests
find_all_v2 in time_entries_repository.py

### DIFF
--- a/tests/time_tracker_api/time_entries/time_entries_query_builder_test.py
+++ b/tests/time_tracker_api/time_entries/time_entries_query_builder_test.py
@@ -4,7 +4,7 @@ from utils.query_builder import CosmosDBQueryBuilder
 from time_tracker_api.time_entries.time_entries_query_builder import (
     TimeEntryQueryBuilder,
 )
-from utils.repository import get_string_without_empty_spaces
+from utils.repository import remove_white_spaces
 
 
 def test_TimeEntryQueryBuilder_is_subclass_CosmosDBQueryBuilder():
@@ -73,9 +73,7 @@ def test_build_with_add_sql_date_range_condition():
                     """
     query = time_entry_query_builder.get_query()
 
-    assert get_string_without_empty_spaces(
-        query
-    ) == get_string_without_empty_spaces(expected_query)
+    assert remove_white_spaces(query) == remove_white_spaces(expected_query)
 
     assert len(time_entry_query_builder.where_conditions) == 1
     assert len(time_entry_query_builder.get_parameters()) == 2
--- a/tests/time_tracker_api/time_entries/time_entries_query_builder_test.py
+++ b/tests/time_tracker_api/time_entries/time_entries_query_builder_test.py
@@ -1,5 +1,4 @@
 import pytest
-from unittest.mock import patch
 from utils.query_builder import CosmosDBQueryBuilder
 from time_tracker_api.time_entries.time_entries_query_builder import (
     TimeEntryQueryBuilder,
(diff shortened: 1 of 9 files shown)


utils/repository.py
```diff
@@ -7,3 +7,10 @@ def convert_list_to_tuple_string(ids_list):
         else str(tuple(ids_list))
     )
     return result
+
+
+def get_string_without_empty_spaces(string: str):
```
Angeluz-07, line 12: This functions seems to be created only for testing the queries. Maybe we could move it to the querybuilder test file. Also consider using the following logic that does the same : `re.sub('[\s]+','', expected_query)`.  Also, it could be called `remove_white_spaces`

utils/query_builder.py
```diff
@@ -0,0 +1,96 @@
...
+                self.parameters.append({'name': f'@{k}', 'value': v})
+        return self
+
+    def add_sql_visibility_condition(self, visible_only: bool):
+        if visible_only:
+            self.where_conditions.append('NOT IS_DEFINED(c.deleted)')
+        return self
+
+    def add_sql_limit_condition(self, limit):
+        if limit and isinstance(limit, int):
```
Angeluz-07, line 44: Not sure if we need the isintance check.

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -196,6 +199,40 @@ def find_all(
             abort(HTTPStatus.NOT_FOUND, "Time entry not found")
         return time_entries
 
+    def find_all_v2(
+        self,
+        event_context: EventContext,
+        owner_id_list: List[str],
```
Angeluz-07, line 205: We also need the pass a list of ids of time-entries.

utils/repository.py
kellycastrof: This functions is called in quer_builder_test and time_entries_query_builder_test 
At the beginning, I had it in both files, but as it would be repeated code I changed it to utils
Do you think it would be better to leave it in both files instead of utils?
I'll rename the function

time_tracker_api/time_entries/time_entries_repository.py
kellycastrof: I can add it

utils/query_builder.py
kellycastrof: What if the person sends a string like "Hello"?
In that case, a syntax error would be generated when executing the query in the database

utils/repository.py
Angeluz-07: Sorry. I see the logic again this `re.sub('[\s]+','', expected_query)` may not work.

Angeluz-07: Try this
```
re.sub("[\s]+",' ', quer).strip()
```

utils/query_builder.py
Angeluz-07: In higher layers we can validate. Or later apply a stronger validation in the whole class. By now I think we are ok with the `if limit`  only. Same for the offset.

tests/time_tracker_api/time_entries/time_entries_query_builder_test.py
```diff
@@ -0,0 +1,78 @@
...
+        (
+            "2021-03-19T05:07:00.000Z",
+            "2021-03-25T10:00:00.000Z",
+            [
+                {"name": "@start_date", "value": "2021-03-19T05:07:00.000Z"},
+                {"name": "@end_date", "value": "2021-03-25T10:00:00.000Z"},
+            ],
+        ),
+        (
+            "2021-02-10T05:08:00.000Z",
```
Angeluz-07, line 29: I have the doubt if the three scenarios are revelant. We may only choose one.

kellycastrof: It could be only one, they are not relevant cases

Angeluz-07 (approved):

eduardisrael (approved):

scastillo-jp (approved):

PR #271, opened 2021-03-29
refactor: TT-182 create method to get all projects with list of ids

eduardisrael:
create method to get all projects with list of ids

### DIFF
--- a/time_tracker_api/projects/projects_model.py
+++ b/time_tracker_api/projects/projects_model.py
@@ -14,7 +14,6 @@ from time_tracker_api.customers.customers_model import (
 from time_tracker_api.customers.customers_model import CustomerCosmosDBModel
 from utils.query_builder import CosmosDBQueryBuilder
 from utils.extend_model import add_customer_name_to_projects
-from utils.repository import create_sql_in_condition
 
 
 class ProjectDao(CRUDDao):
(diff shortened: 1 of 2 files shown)


time_tracker_api/projects/projects_model.py
```diff
@@ -65,24 +65,22 @@ def __init__(self):
...
         )
-            
-    def find_all_with_customer_id_in_list(
+
+    def find_all_v2(
         self,
         event_context: EventContext,
-        customer_ids_list: List[str],
+        ids_list: List[str],
+        customers_ids_list: List[str] = None,
```
kellycastrof, line 73: I think that customer_ids is better, customers_ids_list is too long .

time_tracker_api/projects/projects_model.py
```diff
@@ -124,6 +122,11 @@ def get_all(self, conditions: dict = None, **kwargs) -> list:
         add_customer_name_to_projects(projects, customers)
         return projects
 
+    def get_all_with_id_in_list(self,id_list):
```
kellycastrof, line 125: Maybe it should be discussed if the function will receive only one list because the find_all method supports two lists for sql in conditions.

time_tracker_api/projects/projects_model.py
VanessaIniguez: Thank you for your observation Kelly

time_tracker_api/projects/projects_model.py
```diff
@@ -8,7 +8,7 @@
 from time_tracker_api.database import CRUDDao, APICosmosDBDao
 from typing import List, Callable
 from commons.data_access_layer.database import EventContext
-from utils.repository import convert_list_to_tuple_string, create_sql_in_condition
```
scastillo-jp, line 11: `convert_list_to_tuple_string` is no longer to be used?

kellycastrof: It is still used but in CosmosDBQueryBuilder

eduardisrael: That's right, CosmosDBQueryBuilder uses convert_list_to_tuple_string. It is no longer necessary to import it into projects_model

kellycastrof (approved):

PR #273, opened 2021-04-07
TT-201 validate the current entry and last entry

VanessaIniguez:
Allow update last entry with the new flag 'update_last_entry_if_overlap'


utils/query_builder.py
```diff
@@ -13,6 +19,7 @@ def __init__(self):
         self.where_conditions = []
         self.limit = None
         self.offset = None
+        self.order_by = tuple()
```
Angeluz-07, line 22: I think it should be None.

utils/query_builder.py
```diff
@@ -75,15 +86,24 @@ def __build_limit(self):
         else:
             return ""
 
+    def __build_order_By(self):
```
Angeluz-07, line 89: Why `order_By` ?

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,52 @@ def find_all_v2(
...
+
+        partition_key_value = self.find_partition_key_value(event_context)
+        result = self.container.query_items(
+            query=query_str,
+            parameters=params,
+            partition_key=partition_key_value,
+        )
+
+        function_mapper = self.get_mapper_or_dict(mapper)
+        return list(map(function_mapper, result))
```
Angeluz-07, line 265: The nature of the method is to return one element. I consider we should change the `list(map(...` by using `next`

tests/utils/query_builder_test.py
```diff
@@ -41,7 +41,9 @@ def test_add_select_conditions_should_update_select_list(
     ],
 )
 def test_add_sql_in_condition_should_update_where_list(
-    attribute, ids_list, expected_where_condition_list,
+    attribute,
```
Angeluz-07, line 44: Is this re-formatting expected or the [pre-commit](https://ioetec.slack.com/archives/CUD8VFS86/p1588709542149600) action has not been set?

tests/utils/query_builder_test.py
```diff
@@ -235,3 +261,23 @@ def test_build_with_empty_and_None_attributes_return_query_select_all():
     assert query == expected_query
     assert len(query_builder.get_parameters()) == 0
     assert len(query_builder.where_conditions) == 0
+
+
+def test_order_by_condition():
```
Angeluz-07, line 266: Please. Make a parametrized version of this method. It seems is possible.

Angeluz-07: Also, I think the method name should be `test_add_sql_order_by_condition`

tests/utils/query_builder_test.py
```diff
@@ -235,3 +261,23 @@ def test_build_with_empty_and_None_attributes_return_query_select_all():
...
+    query_builder = CosmosDBQueryBuilder().add_sql_order_by_condition(
+        'start_date', Order.DESC
+    )
+
+    assert len(query_builder.order_by) == 2
+    assert query_builder.order_by == ('start_date', 'DESC')
+
+
+def test__build_orderBy():
```
Angeluz-07, line 275: Here, I think it is also possible to make a parametrized version.

time_tracker_api/api.py
```diff
@@ -34,6 +34,19 @@ def remove_required_constraint(model: Model):
...
+def add_update_last_entry_if_overlap(time_entry_model: Model):
+    time_entry_flag = {
+        'update_last_entry_if_overlap': fields.Boolean(
+            title='Update last entry if overlap',
+            required=False,
+            description='Flag that indicates if the last time entry is updated',
+            example=True,
+        )
+    }
+    new_model = time_entry_model.clone('TimeEntryInput', time_entry_flag)
```
Angeluz-07, line 46: It is neccesary to use the `clone` method?. Because if not,the method woudl be simpler.

time_tracker_api/time_entries/time_entries_dao.py
```diff
@@ -226,6 +226,9 @@ def create(self, data: dict):
     def update(self, id, data: dict, description=None):
         event_ctx = self.create_event_context("update", description)
 
+        if data.get('update_last_entry_if_overlap'):
```
Angeluz-07, line 229: I think, this get, should have a False default value, to make it more robust.

Angeluz-07: Also, this code should be after the line `self.check_whether_current_user_owns_item(time_entry)`

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,52 @@ def find_all_v2(
...
+        owner_id = data.get('owner_id')
+        last_entry = self.get_last_entry(owner_id, event_context)
+        last_entry = last_entry.pop()
+
+        end_date = str_to_datetime(last_entry.end_date)
+        start_date = data.get('start_date')
+        start_date = str_to_datetime(start_date)
+
+        if start_date < end_date:
+            update_date = {'end_date': data.get('start_date')}
```
Angeluz-07, line 277: `update_date` by `new_end_date`

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,52 @@ def find_all_v2(
...
+            return self.partial_update(
```
Angeluz-07, line 278: I think there is no need to return here as the return value is not being used.

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,52 @@ def find_all_v2(
...
+                last_entry.id, update_date, event_context
+            )
+
+        return False
```
Angeluz-07, line 282: No need to return. But if you want, you can return None explicitly.

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,52 @@ def find_all_v2(
...
+        event_context: EventContext,
+        visible_only=True,
+        mapper: Callable = None,
+    ):
+        query_builder = (
+            CosmosDBQueryBuilder()
+            .add_sql_where_equal_condition({'owner_id': owner_id})
+            .add_sql_order_by_condition('end_date', Order.DESC)
+            .add_sql_limit_condition(1)
+            .add_sql_offset_condition(1)
```
Angeluz-07, line 250: Could you please evaluate if you really need the offset condition.

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,52 @@ def find_all_v2(
...
+        result = self.container.query_items(
+            query=query_str,
+            parameters=params,
+            partition_key=partition_key_value,
+        )
+
+        function_mapper = self.get_mapper_or_dict(mapper)
+        return list(map(function_mapper, result))
+
+    def update_last_entry(self, data: dict, event_context: EventContext):
```
Angeluz-07, line 267: I would suggest don deal with `data` dict here. Pass the parameters you only need directly. owner_id, start_date.

time_tracker_api/time_entries/time_entries_repository.py
eduardisrael: You could change like this next(iterable, default) -> function_mapper(next(result)), maybe handle an exception only if necessary

utils/query_builder.py
kellycastrof: It was a typing error

time_tracker_api/time_entries/time_entries_repository.py
eduardisrael, line 265: You could change like this next(iterable, default) -> function_mapper(next(result)), maybe handle an exception only if necessary

tests/utils/query_builder_test.py
kellycastrof: It was the pre-commit action

time_tracker_api/api.py
kellycastrof: It is necessary, the idea was used extended method, but it was deprecated. 
I didn't see the way to add a new field without changing the TimeEntry model defined in time_entries_namespace

time_tracker_api/time_entries/time_entries_repository.py
kellycastrof: Yeah, it will be better only does the partial update
With that, the method return None by default

time_tracker_api/time_entries/time_entries_repository.py
kellycastrof: It is needed to take the last entry instead of getting the running time entry. We evaluate your suggestion of excluding the running time entry id, but it was necessary to create other methods to carry out that. 
So, we thought it was better to use a function that we already have.

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,48 @@ def find_all_v2(
...
+            query=query_str,
+            parameters=params,
+            partition_key=partition_key_value,
+        )
+
+        function_mapper = self.get_mapper_or_dict(mapper)
+        return function_mapper(next(result))
+
+    def update_last_entry(
+        self, owner_id, start_date, event_context: EventContext
```
Angeluz-07, line 268: could you please add type hinting for owner_id and start_date.

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -233,6 +235,48 @@ def find_all_v2(
...
+    ):
+        last_entry = self.get_last_entry(owner_id, event_context)
+        start_date_tmp = start_date
+
+        end_date = str_to_datetime(last_entry.end_date)
+        start_date = str_to_datetime(start_date)
```
Angeluz-07, line 274: I think you meant `str_to_datetime(start_date_tmp)`.

time_tracker_api/time_entries/time_entries_dao.py
```diff
@@ -225,10 +225,14 @@ def create(self, data: dict):
 
     def update(self, id, data: dict, description=None):
         event_ctx = self.create_event_context("update", description)
-
         time_entry = self.repository.find(id, event_ctx)
         self.check_whether_current_user_owns_item(time_entry)
 
+        if data.get('update_last_entry_if_overlap'):
```
Angeluz-07, line 231: `data.get('update_last_entry_if_overlap', None)` or `data.get('update_last_entry_if_overlap', False)`

time_tracker_api/time_entries/time_entries_repository.py
kellycastrof: No, start_date or start_date_tmp it's ok, but with line 274 I convert the string in a DateTime object, so it causes an error if a pass star_date in the update_data because it is not a string, I create the start_dte_tmp to keep that string

time_tracker_api/time_entries/time_entries_dao.py
kellycastrof: get method return None by default without passing default value, but I can add None if you think it is necessary.

time_tracker_api/time_entries/time_entries_repository.py
Angeluz-07: You can use `_start_date` variable name, so you avoid the use of `start_date_tmp` . Choose what it works best for you.

Angeluz-07 (approved):
Very good job.

eduardisrael (approved):

PR #274, opened 2021-04-09
fix: TT-206 return 404 when a time entry is not running

eduardisrael:
Handle this error with a message 204

### DIFF
--- a/time_tracker_api/api.py
+++ b/time_tracker_api/api.py
@@ -2,7 +2,6 @@ from azure.cosmos.exceptions import (
     CosmosResourceExistsError,
     CosmosResourceNotFoundError,
     CosmosHttpResponseError,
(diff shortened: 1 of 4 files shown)


time_tracker_api/api.py
```diff
@@ -2,6 +2,7 @@
     CosmosResourceExistsError,
     CosmosResourceNotFoundError,
     CosmosHttpResponseError,
+    CosmosClientTimeoutError,
```
scastillo-jp, line 5: I have a question,  How does this help?

eduardisrael: I thought of handling it as 'An operation failed to complete within the specified timeout.' However, it is not necessary for a 204 message. Thanks for asking.

scastillo-jp (approved):

PR #275, opened 2021-04-14
fix:TT-210 Dont allow activities deleting

pafasosacoder:


commons/data_access_layer/cosmos_db.py
```diff
@@ -133,6 +133,13 @@ def create_sql_condition_for_visibility(
             return 'AND NOT IS_DEFINED(%s.deleted)' % container_name
         return ''
 
+    @staticmethod
+    def create_sql_condition_for_active( only_active: bool,
+        container_name='c' ) -> str:
+        if only_active:
+            return 'AND NOT IS_DEFINED(%s.status) OR (IS_DEFINED(%s.status) AND %s.status = \'active\')' % (container_name, container_name, container_name)
```
kellycastrof, line 140: This line is too long, can you consider put all inside () and add an Enter after '%' character? 👀

commons/data_access_layer/cosmos_db.py
```diff
@@ -133,6 +133,13 @@ def create_sql_condition_for_visibility(
             return 'AND NOT IS_DEFINED(%s.deleted)' % container_name
         return ''
 
+    @staticmethod
+    def create_sql_condition_for_active( only_active: bool,
```
scastillo-jp, line 137: rename  **create_sql_condition_for_active** to **active_condition**

Angeluz-07: I think `create_sql_is_active_condition` would be ok to keep the standard naming present in the code.

commons/data_access_layer/cosmos_db.py
Angeluz-07: Try using the .format() method instead for readability.
 ex.
```
tool="Unsupervised algorithms"
goal="patterns"
print("{title} try to find {aim} in the dataset".format(title=tool, aim=goal))
```

commons/data_access_layer/cosmos_db.py
```diff
@@ -225,6 +232,7 @@ def find(
     def find_all(
         self,
         event_context: EventContext,
+        only_active = True,
```
Angeluz-07, line 235: Comparing this with `visible_only` parameter. It would be good to rename this var to `active_only`.

commons/data_access_layer/cosmos_db.py
```diff
@@ -348,11 +358,11 @@ class CosmosDBDao(CRUDDao):
     def __init__(self, repository: CosmosDBRepository):
         self.repository = repository
 
-    def get_all(self, conditions: dict = None, **kwargs) -> list:
+    def get_all(self, only_active = True, conditions: dict = None, **kwargs) -> list:
```
Angeluz-07, line 361: I think is not neccesary to put this value here(unless you plan to do something with it). Is already defined in the repository method.

commons/data_access_layer/cosmos_db.py
pafasosacoder: Ok, It has been adjusted, using a new approach for avoiding the only_active argument.

commons/data_access_layer/cosmos_db.py
pafasosacoder: Ok, it has been adjusted, with a new approach to avoid the only_active argument.

commons/data_access_layer/cosmos_db.py
pafasosacoder: Ok, it has been adjusted, with a new approach to avoid the only_active argument.

commons/data_access_layer/cosmos_db.py
pafasosacoder: Ok, it has been adjusted, with a new approach to avoid the only_active argument.

commons/data_access_layer/cosmos_db.py
```diff
@@ -133,6 +133,30 @@ def create_sql_condition_for_visibility(
...
+            condition_operand = ' AND '
+            if status_value == 'active':
+                not_defined_condition = (
+                    'AND NOT IS_DEFINED({container_name}.status)'.format(
+                        container_name=container_name
+                    )
+                )
+                condition_operand = ' OR '
+
+            defined_condition = '(IS_DEFINED({container_name}.status) AND {container_name}.status = \'{status_value}\')'.format(
```
israteneda, line 151: Maybe we can split this line for a better readability and to follow the [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) specification.

scastillo-jp (approved):

PR #281, opened 2021-04-23
fix: TT-225 Add build query with not in

VanessaIniguez:
Add build query with not in and add id_running_entry in params

### DIFF
--- a/time_tracker_api/time_entries/time_entries_repository.py
+++ b/time_tracker_api/time_entries/time_entries_repository.py
@@ -248,7 +248,7 @@ class TimeEntryCosmosDBRepository(CosmosDBRepository):
             .add_sql_where_equal_condition({'owner_id': owner_id})
             .add_sql_order_by_condition('end_date', Order.DESC)
             .add_sql_not_in_condition('id', [id_running_entry])
-            .build_with_order_by()
+            .build()
         )
         query_str = query_builder.get_query()
         params = query_builder.get_parameters()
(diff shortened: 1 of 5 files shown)


time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -246,29 +247,26 @@ def get_last_entry(
             CosmosDBQueryBuilder()
             .add_sql_where_equal_condition({'owner_id': owner_id})
             .add_sql_order_by_condition('end_date', Order.DESC)
-            .add_sql_limit_condition(1)
-            .add_sql_offset_condition(1)
-            .build()
+            .add_sql_not_in_condition('id', [id_running_entry])
+            .build_with_order_by()
```
kellycastrof, line 251: Why are there now two methods for the build?
What is the difference?

utils/query_builder.py
```diff
@@ -109,6 +117,18 @@ def build(self):
         )
         return self
 
+    def build_with_order_by(self):
```
kellycastrof, line 120: In this method you do the same as in the build method, it would not be necessary to add it since in this way we would have a build method for each combination when the general build method already does this.
Or did the build method generate an error?

utils/query_builder.py
```diff
@@ -61,6 +62,13 @@ def add_sql_order_by_condition(self, attribute: str, order: Order):
         self.order_by = (attribute, order.name)
         return self
 
+
+    def add_sql_not_in_condition(self, attribute: str = None, ids_list: List[str] = None):
```
kellycastrof, line 66: There is no test method for this new condition in query_builder_test

scastillo-jp (approved):

PR #282, opened 2021-04-27
fix: TT-219 Refactoring find_all for time entries

pafasosacoder:


tests/time_tracker_api/time_entries/time_entries_namespace_test.py
```diff
@@ -204,18 +214,34 @@ def test_get_time_entry_should_succeed_with_valid_id(
...
 @patch('utils.azure_users.AzureConnection.get_token', Mock())
 @patch('utils.azure_users.AzureConnection.is_test_user')
 @patch('utils.azure_users.AzureConnection.get_test_user_ids')
 @pytest.mark.parametrize(
     'current_user_is_tester, expected_user_ids',
     [
-        (True, ['id1', 'id2']),
-        (False, ['id2']),
+        (True, ['id1', 'id1']),
+        (False, ['id1', 'id1']),
```
israteneda, line 233: just a question why we have the `ids` with the same value

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -141,7 +141,37 @@ def count(
...
+            activities = activity_dao.get_all(
+                custom_sql_conditions=[custom_conditions_activity],
+                visible_only=False,
+                max_count=max_count,
+            )
+            add_activity_name_to_time_entries(time_entries, activities)
+
+            users = AzureConnection().users()
+            add_user_email_to_time_entries(time_entries, users)
+        elif not time_entries and len(conditions) > 1:
```
israteneda, line 170: just a question why don't use a `if` instead of `elif`

time_tracker_api/time_entries/time_entries_repository.py
```diff
@@ -199,41 +229,71 @@ def find_all(
...
         function_mapper = self.get_mapper_or_dict(mapper)
-        return list(map(function_mapper, result))
+        time_entries = list(map(function_mapper, result))
+
+        return self.add_complementary_info(time_entries, max_count)
+
+    def create_sql_test_users_exclusion_condition(self, test_user_ids):
+        if test_user_ids != None:
+            return (
+                " AND c.owner_id NOT IN ('" + "','".join(test_user_ids) + "') "
```
israteneda, line 294: This statement seems a bit complex, do you think it can be simplified with a format string, something like that:
```
" AND c.owner_id NOT IN ('{}') ".format("','".join(test_user_ids))
```
just a suggestion

kellycastrof, line 294: It's hard to understand, did you consider using the convert_list_to_tuple_string function?

tests/time_tracker_api/time_entries/time_entries_model_test.py
```diff
@@ -295,21 +324,29 @@ def test_find_all_v2(
     time_entry_repository.container = Mock()
     time_entry_repository.container.query_items = query_items_mock
 
-    result = time_entry_repository.find_all_v2(
-        event_context,
-        ['owner_id'],
-        {
+    time_entry_repository.add_complementary_info = Mock()
+    time_entry_repository.add_complementary_info = query_items_mock
```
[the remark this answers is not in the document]
pafasosacoder: Ok, the line 327 has been removed.

tests/time_tracker_api/time_entries/time_entries_namespace_test.py
pafasosacoder: Ok, the test was split in two tests because there is a need to verify when method fetches time entries for test user and for no test user.

time_tracker_api/time_entries/time_entries_repository.py
pafasosacoder: That conditional, besides it is mutually exclusive, has another condition, so I recommend keeping the same verification.

time_tracker_api/time_entries/time_entries_repository.py
pafasosacoder: Ok, the expression was replaced for the use of convert_list_to_tuple_string generic function

pafasosacoder: Ok, the expression was replaced for the use of convert_list_to_tuple_string generic function

scastillo-jp (approved):

PR #295, opened 2021-06-01
fix: TT-257 customer name null on get project by id

jcalarcon98:
## Problem

There is an error when trying to consulting to the endpoint:

```bash
<time-tracker-backend-url>/projects/<project-id>
```
The response brings the customer name of the project **null**.

## Solution

The method **get** on **ProjectCosmosDBDao** class has been modified and add the correct behavior to fix this error.

Also, tests have been added to verify its correct operation.


kevinjlope (approved):
good job!

PR #301, opened 2021-06-18
fix: TT-270 Handle exceptions in create update entries

jcalarcon98:
## Description
Currently, if we want to create a new entry and we don't pass `activity_id` field, the  activity will be create without a problem. 

For example: 

image

This behavior can cause problems when we try to get all the Entries and in the UI part, it will not render reports Page, and time Entries page.

With  this PR, it is checked that an activity and a project are provided, and also that the projects or activities entered exist. for example:
**Entry without `activity_id` field:**

image

**Entry with fake `activity_id` field: **
image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 import copy
 from datetime import datetime, timedelta
+from http import HTTPStatus
 
 import jwt
 import pytest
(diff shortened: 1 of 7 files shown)


kevinjlope (approved):
Good job!

scastillo-jp (approved):

PR #305, opened 2021-07-07
fix: TT-287 Fix users email doesn't show in reports page

jcalarcon98:
## Problem Description:

Currently the Reports Page does not load the emails of some users, this is because the current implementation does not take into account that the Azure endpoint only returns the first 100 users.
image

##  Solution:

In the user request, if there are more users, the Azure api returns an additional field named `o.datanextLink`, as long as the API returns this field he following number of remaining users is consumed.
image

### DIFF
--- a/tests/utils/azure_users_test.py
+++ b/tests/utils/azure_users_test.py
@@ -1,4 +1,7 @@
+import copy
 from unittest.mock import Mock, patch
+from requests import Response
+
 from utils.azure_users import AzureConnection, ROLE_FIELD_VALUES, AzureUser
 from pytest import mark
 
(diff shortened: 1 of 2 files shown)


tests/utils/azure_users_test.py
```diff
@@ -228,3 +242,26 @@ def test_remove_user_from_group(
...
+
+    second_response = copy.copy(first_response)
+    second_response._content = b'{"value":[{"displayName":"Fake2","otherMails":["fake2@ioet.com"],"objectId":"1"}]}'
+
+    get_mock.side_effect = [first_response, second_response]
+    get_groups_and_users_mock.return_value = []
+
+    users = AzureConnection().users()
+
+    assert len(users) == 2
```
jase156, line 267: In my opinion, I think that an assert should be added to compare if the response's elements are the same as the fake.

`users == get_mock.side_efect`

jcalarcon98: Resolved in an internal discussion

jase156 (approved):
Good Job Jean Carlos

PieritoAlva95 (approved):
Very good job Jean Carlos, keep up the good work. 💯 👍🏼 🥇

scastillo-jp (approved):

PR #307, opened 2021-07-08
fix: TT-274 fix error on archive a customer removes the project

PieritoAlva95:
## Problem Description:
Currently, when archiving a customer, the columns of projects and customers cannot be displayed in the time entries and report pages.
Screenshot from 2021-07-08 11-48-47

## Solution:
The code was modified so that when a customer is archived it does not affect the functionality of the time entries and report pages, and the columns with problems can be displayed normally.
Screenshot from 2021-07-08 11-53-51

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,5 @@ env.*
 
 # SSL certificate for cosmos emulator
 emulatorcert.crt
+
+seed_database.json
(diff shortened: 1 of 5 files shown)


jcalarcon98 (approved):
Good Job Jean 💥

scastillo-jp (approved):

PR #308, opened 2021-07-12
fix: TT-288 Fix and improve tests on find_interception_with_date_range function

jcalarcon98:
## Description
Currently there are tests that validate the operation of the find_interception_with_date_range function, however, these tests are creating entries in the production database, it is also important to clarify that each of the tests deletes the entry created.

## Solution

To prevent these entries from being created, it was decided to mock the return value when querying entries that clash with the entry to be created


PieritoAlva95 (approved):
Very good work Jean Carlos. 👍🏼 💯 🥇

kevinjlope (approved):

jase156 (approved):
Good job Jean Carlos

PR #319, opened 2021-09-03
feat: TT-326 Get recent projects

jcalarcon98:
## Description
Currently there is no endpoint that allows to know which are the latest projects in which the user has created entries. This functionality is performed by an external library, which stores the projects in the local storage browser. With this PR this functionality has been created to extract the projects in which the user has generated entries, taking into account the user's last 20 entries.


JosueOb (approved):
I have finished reviewing the changes made and the functionality. So, I have performed tests on the endpoint /projects/recent, which returns the projects used in the last 20 entries, using the test and production database, and in both cases, the projects used have been returned, and I have not had any problems. Good work @jcalarcon98.

PR #325, opened 2021-10-11
feat: TT-356 Read activities with an azure endpoint

EdansRocks:
With this PR an Azure Endpoint was created using Azure Functions according to the new backend architecture for the Time Tracker project to read the activities data using the use cases created on (TT-352):
image

### DIFF
--- a/V2/azure_time_tracker/time_tracker/tests/api/api_fixtures.py
+++ b/V2/azure_time_tracker/time_tracker/tests/api/api_fixtures.py
@@ -1,0 +1,44 @@
+import json
+import pytest
+import shutil
+
+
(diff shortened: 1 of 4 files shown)


V2/azure_time_tracker/host.json
```diff
@@ -0,0 +1,15 @@
+{
```
scastillo-jp, line 1: what is the purpose of using this host.json file

EdansRocks: It's a config file created by Azure when you create a new project that contains the configuration options that affect all functions in a function app instance, remember this file will be replaced by the host.json of serverless.

PR #328, opened 2021-10-14
TT-358 Use serverless to create Azure endpoint

EdansRocks:
With this PR an Azure Endpoint was created using serverless according to the new backend architecture for the Time Tracker project to read the activities data using the use cases created on (TT-352). The structure based on Azure functions (TT-356) was replaced with serverless framework ([click here](https://www.serverless.com) for more info).
 
image

### DIFF
--- a/V2/Makefile
+++ b/V2/Makefile
@@ -5,11 +5,3 @@ install:
 		pip install -r requirements.txt
 		@echo "Completed! "
 
-deploy:
-		@echo "Using serverless.yml provider config to deploy"
-		@echo "Deploying Time Tracker to Azure..."
-		sls deploy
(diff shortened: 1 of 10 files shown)


V2/Makefile
```diff
@@ -0,0 +1,15 @@
+install:
+		@echo "Installing Time Tracker"
+		npm install
+		pip install --upgrade pip
+		pip install -r requirements.txt
+		@echo "Completed! "
+
+deploy:
```
faustocv, line 8: does it deploy to a upper or lower environment?

eguezgustavo: at this point it deploys to prod

PR #339, opened 2021-11-10
fix: TT-393 userid convert to list

mandres2015:
Add more users to admin group, actually we are using a simple filed string, each id es separated with ","


utils/azure_users.py
```diff
@@ -263,7 +263,8 @@ def get_groups_and_users(self):
             [member['objectId'] for member in item['members']],
         )
         result = list(map(parse_item, response.json()['value']))
-        result[0][1].append(self.config.USERID)
+        list_of_userid = self.config.USERID.split(" ")
```
scastillo-jp, line 266: - change ** list_of_userid ** to ** users_id ** 
- I recommend using `,` as the split symbol.

dsgarcia8: Resolve comment

scastillo-jp (approved):

PR #350, opened 2021-11-16
feat: TT-399 Config use makefile to executing tests

gancoce94:
PR - Config Makefile for automate tests
Execute 'make'
```
---------------HELP-----------------
To install the dependencies type make install
To test the project type make test
To run the local database type make start-local
To run all comands type make ci
------------------------------------
```

### DIFF
--- a/V2/Makefile
+++ b/V2/Makefile
@@ -16,6 +16,8 @@ install:
 		@echo "Completed! "
 
 .PHONY: test
+test: export ENVIRONMENT = test
+test: export TEST_DB_CONNECTION = sqlite:///:memory:
 test:
 		@echo "=========================================Lint with flake8========================================="
 		flake8 . --show-source --statistics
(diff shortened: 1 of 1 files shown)


V2/Makefile
```diff
@@ -1,8 +1,31 @@
...
+		@echo "To run all comands type make ci"
+		@echo "------------------------------------"
+
+.PHONY: install
 install:
-		@echo "Installing Time Tracker"
+		@echo "=========================================Installing dependencies Time Tracker========================================="
 		npm install
-		pip install --upgrade pip
+		# pip install --upgrade pip
```
RobertoCordova, line 14: Question: Why do you comment on this line? Don't we want to update?

RobertoCordova (changes requested):

V2/Makefile
gancoce94: In some laptop pip upgrade dont func, so that was de the reason, in the newest commit is solved.

RobertoCordova (approved):

RobertoCordova (approved):

scastillo-jp (approved):

PR #353, opened 2021-11-18
feat:TT-407 list latest v2 time entries

mandres2015:
This pull request get the n (20 default) latest entries from an user

### DIFF
--- a/V2/tests/unit/use_cases/time_entries_use_case_test.py
+++ b/V2/tests/unit/use_cases/time_entries_use_case_test.py
@@ -75,3 +75,16 @@ def test__get_time_entry_by_id_function__uses_the_time_entry_service__to_retriev
 
     assert time_entry_service.get_by_id.called
     assert expected_time_entries == actual_time_entry
+
+
+def test__get_latest_entries_function__uses_the_time_entry_service__to_get_last_entries(
+    mocker: MockFixture,
+):
+    expected_latest_time_entries = mocker.Mock()
+    time_entry_service = mocker.Mock(get_latest_entries=mocker.Mock(return_value=expected_latest_time_entries))
+
+    time_entry_use_case = _use_cases.GetLastestTimeEntryUseCase(time_entry_service)
+    latest_time_entries = time_entry_use_case.get_latest_entries(Faker().pyint(), Faker().pyint())
+
+    assert time_entry_service.get_latest_entries.called
+    assert expected_latest_time_entries == latest_time_entries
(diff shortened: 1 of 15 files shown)


V2/time_tracker/time_entries/_infrastructure/_data_persistence/_time_entries_dao.py
```diff
@@ -1,6 +1,6 @@
 import dataclasses
-
 import sqlalchemy
```
josepato87, line 2: Maybe we could import sqlachemy as sq, so we can avoid typing all the word.

V2/time_tracker/time_entries/_application/_time_entries/_get_latest_entries.py
```diff
@@ -0,0 +1,47 @@
...
+
+    try:
+        owner_id = req.params.get("owner_id")
+        limit = req.params.get("limit")
+
+        if not owner_id:
+            return func.HttpResponse(
+                body="No owner id found",
+                status_code=404,
+                mimetype="application/json"
```
josepato87, line 24: We could create a constants file for server responses and errors instead of hardcoding the strings

V2/tests/unit/use_cases/time_entries_use_case_test.py
```diff
@@ -30,3 +30,16 @@ def test__delete_time_entry_function__uses_the_time_entry_service__to_delete_tim
 
     assert time_entry_service.delete.called
     assert expected_time_entry == deleted_time_entry
+
+
+def test__get_latest_entries_function__uses_the_time_entry_service__to_get_last_entries(
```
josepato87, line 35: "function" in the test name is redundant

josepato87 (changes requested):
Left some comments

V2/tests/unit/use_cases/time_entries_use_case_test.py
mandres2015: We are going to create a ticket, this or next week to make these changes and standardize the code to avoid these possible errors when writing code, so I do not make the change in this PR

V2/time_tracker/time_entries/_application/_time_entries/_get_latest_entries.py
mandres2015: We are going to create a ticket, this or next week to make these changes and standardize the code to avoid these possible errors when writing code, so I do not make the change in this pr

scastillo-jp (approved):

PR #356, opened 2021-11-19
ci: TT-411 adds pull request (ci) workflow for time-tracker v1

cristiantoaquizadev:
Since our Azure DevOps pipeline has no more minutes remaining to be run, we have the need to have a new way to run the continuous integration process. That way is using GitHub Actions and this PR is to add the workflow to run some steps when a pull request is created.


josepato87 (approved):
LGTM

PR #369, opened 2022-03-09
Fix: All ioet users are returned from AD

cxcarvaj:
* Fix get all users from AD.
* Fix dependencies errors.
* Fix tests due to changes in the code.

### DIFF
--- a/tests/utils/azure_users_test.py
+++ b/tests/utils/azure_users_test.py
@@ -264,4 +264,4 @@ def test_users_functions_should_returns_all_users(
 
     users = AzureConnection().users()
 
-    assert len(users) == 0
+    assert len(users) == 2
--- a/utils/azure_users.py
+++ b/utils/azure_users.py
@@ -133,9 +133,10 @@ class AzureConnection:
                 )[1]
                 endpoint = endpoint + skip_token_attribute + request_token
                 
-        for i in range(len(users)):
-            if users[i]['mail'] is None:
-                valid_users.append(users[i])
+        for user in users:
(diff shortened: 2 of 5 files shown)


utils/azure_users.py
```diff
@@ -133,9 +133,10 @@ def users(self) -> List[AzureUser]:
                 )[1]
                 endpoint = endpoint + skip_token_attribute + request_token
                 
-        for i in range(len(users)):
-            if users[i]['mail'] is None:
-                valid_users.append(users[i])
+        for user in users:
+            user_emails = user['otherMails']
+            if(len(user_emails) != 0 and user_emails[0].split('@')[1] == 'ioet.com'):
```
FranLMSP, line 138: Maybe we could put `user_emails[0].split('@')[1]` into a variable `email_domain`? also why is it only checking for the first one?

cxcarvaj: Because Azure returns that as a list and always will be just one email, and sure we can change that into a variable!

LuisMS7 (approved):

tests/utils/azure_users_test.py
```diff
@@ -264,4 +264,4 @@ def test_users_functions_should_returns_all_users(
 
     users = AzureConnection().users()
 
-    assert len(users) == 0
+    assert len(users) == 2
```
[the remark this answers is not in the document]
cxcarvaj: Actually yes we can, but we will keep that as it is because its not actually the users in the db, they are mock users

